PR #15, opened 2015-02-03
Replace oembed partial w/ a partial that will either render the standard...

jkeck:
... OSD viewer for the uploaded resource or load the PURL Embed.

Just a thought on how to handle this in a dead-simple way.  Not sure if you had a more elegant solution in mind.


cbeer:
Yeah, this'll work for now. When we have a third type of display, we should figure out the better solution (using document types? i don't know..)

PR #34, opened 2015-10-13
Pin to Rails 4.2.2 to fix regression in mounted engine routing

cbeer:


ndushay:
this looks good to me.  shall I merge?

PR #47, opened 2015-10-15
Expose MODS xml as a blacklight document export format; fixes #30

### LINKED ISSUE #30: Expose MODS XML as a Blacklight document export format

cbeer:


config/initializers/mime_types.rb
```diff
@@ -2,4 +2,4 @@
 
 # Add new mime types for use in respond_to blocks:
 # Mime::Type.register "text/richtext", :rtf
-# Mime::Type.register_alias "text/html", :iphone
+Mime::Type.register_alias "text/xml", :mods
```
ndushay: is there any risk if we have another xml format added in the future?  or is this just saying that asdfas.mods  will have mime type  text/xml   and    asdasdfas.foo  can also have mime type text/xml?

cbeer: Yep, that's it.

PR #56, opened 2015-10-21
Add donor tags to the catalog controller configuration

peetucket:
adds the donor tag facet to blacklight config
adds hound and rubocop
fixes lint violations
updates readme and adds sample gdor.yml file to get things working in development
updates dependencies in gemfile and eliminates "remove-oai" branch references for harvestdor gem

@cbeer @ndushay


app/controllers/catalog_controller.rb
```diff
@@ -69,26 +69,27 @@ class CatalogController < ApplicationController
...
-    config.add_facet_field 'pub_date', label: 'Date'  
+    config.add_facet_field 'pub_date', label: 'Date'
     config.add_facet_field 'language', label: 'Language'
     config.add_facet_field 'author_person_facet', label: 'Author', limit: true
     config.add_facet_field 'topic_facet', label: 'Topic', limit: true
     config.add_facet_field 'geographic_facet', label: 'Region', limit: true
-    config.add_facet_field 'era_facet', label: 'Era'  
+    config.add_facet_field 'era_facet', label: 'Era'
     config.add_facet_field 'author_other_facet', label: 'Organization (as author)', limit: true
+    config.add_facet_field 'donor_tags_sim', label: 'Donor tags', limit: true
```
ndushay: I might like a comment in here or somewhere to indicate
1.  it only applies to Feigenbaum as of 2015/10
2.  perhaps this shouldn't display for ALL exhibits -- i think it will be noisy for curators (and they might think "hey, can't I have donor tags)
3.  in the future we may refactor so that single exhibit specific configuration is possible

peetucket: will add comments in next set of work

PR #72, opened 2015-10-30
Try sucker_punch as an ActiveJob (background-processing) adapter

cbeer:
As noted in the config:

```
  # Sucker Punch is a single-process Ruby asynchronous processing library.
  # The benefit to this is there is no additional infrastructure requirement
  # (ie. database, redis, etc.). The downside is that if the web processes is
  # restarted and there are jobs that haven't yet been processed, they will be lost.
  #
  # Spotlight only uses this for data processing that ought to be user-restartable
  # A different queue adapter would be preferable, but requires some significant
  # thought about how to do that in our single-tenant/many-application environment
```


ndushay:
I regret to say I don't know what problem this is trying to solve, or which users would be doing what to which queues.  End users refreshing pages?  admin users restarting indexing?   Is this about Peter's random 500 errors spinning up the client locally?  Or about the really long waits when we go to /requests page?

cbeer:
By default, ActiveJob uses a synchronous queue adapter. Spotlight uses ActiveJob for indexing tasks, so hopefully, after spinning it out to a different thread, the "Reindex" and "Add item" pages will return quickly and processing will happen in the background.

ndushay:
:+1:

ndushay:
have we used this anywhere else?  Is there any risk involved?  I don't want to deploy while you're away and fooch prod exhibits.

cbeer:
Never, this the first I've even heard of it. Nothing it could break is user facing (just curator tools), and none of our non-DLSS curators probably even use this functionality. If it happens to break, it should be nice and easy to revert.

PR #76, opened 2015-11-02
update schema.xml so that jetty solr has correct multivalued field co…

peetucket:
…nfig for full text search

this makes the local jetty match what is happening in production

after doing this you need to do a "rake spotlight:configure_jetty" and then restart your jetty

connects to sul-dlss/spotlight-dor-resources#2


ndushay:
Peter, I think the email I just sent may inform what we do here.   I'm going to hold off on this for now.   I think your changes to spotlight-dor-resources may be more crucial, if the production solr configs are ok??

indexing email

ndushay:
will address this need with new approach for solr-configs.  PR coming your way soon.

PR #79, opened 2015-11-05
update to latest version of spotlight-dor-resources with full text in…

peetucket:
…dexing

@ndushay 

connects to sul-dlss/spotlight-dor-resources#2

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -117,7 +117,7 @@ GEM
     bundler-audit (0.4.0)
       bundler (~> 1.2)
       thor (~> 0.18)
-    byebug (6.0.2)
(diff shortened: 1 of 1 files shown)


Gemfile.lock
```diff
@@ -467,7 +467,7 @@ GEM
       twitter-text (~> 1.4)
     sitemap_generator (5.1.0)
       builder
-    social-share-button (0.1.8)
+    social-share-button (0.1.10)
```
cbeer: Did you mean to update `social-share-button`? It's probably not a problem, but...

peetucket: did a bundle update overall, accepted that update, though probably not necessary in any way

PR #84, opened 2015-11-11
add full text restricted search to drop-down fielded search menu options

peetucket:
@ndushay @cbeer 

connects to #2

### DIFF
--- a/app/controllers/catalog_controller.rb
+++ b/app/controllers/catalog_controller.rb
@@ -196,6 +196,14 @@ class CatalogController < ApplicationController
       }
     end
 
+    config.add_search_field('full_text') do |field|
+      field.label = 'Full text'
+      field.include_in_advanced_search = false
+      field.solr_parameters = { defType: 'lucene' }
+      field.solr_local_parameters = {
+        df: 'full_text_tesim'
+      }
+    end
     # "sort results by" select (pulldown)
     # label in pulldown is followed by the name of the SOLR field to sort by and
     # whether the sort is ascending or descending (it must be asc or desc


ndushay:
Peter - are you sure that this searches the right thing in solr when you select it?

peetucket:
i did a test by searching for a very specific phrase only found in the full text and it returned the result when restricted to the full text field but not when restricted to one of the other specific fields (of course it also returned with "all fields")

cbeer:
I wonder what the value of searching just the (very messy) full text is. I wonder if it'd be better for our users if we added it to the "all fields" parameters with appropriate relevancy tuning.

peetucket:
It does already return results from "all fields", though additional relevancy tuning could be good.  Having it as an option seems interesting to me, even if the OCR results are messy (and they asked for it as well in this case). It is of course easily turned off per exhibit so it doesn't have to pollute other sites.  Perhaps other full text projects will have better OCR as well?

ndushay:
perhaps this is a design question?

Once I get them in, the new changes to the solr configs separate full text into its own Solr fields, separate from metadata (e.g. all_search), but it *is* searched as part of the generic search.

cbeer:
@ggeisler ?

ggeisler:
I think it's potentially useful for the user to be able to restrict a search to full-text (I can imagine a scenario where a user wants to find documents where a specific term or phrase is found only in the document text and not have documents returned that only have that term or phrase in the, for example, title field).

That scenario might not realistically come up with Salt specifically, but it seems possible with another collection, especially, as @peetucket said, if it had better OCR. 

So I agree restricting by full-text might not be particularly helpful for users of Salt, but the feature in general could be useful, and it might not cause a problem (even if not adding value over an all fields search) for the Salt user that chooses to use it?

cbeer:
@peetucket can you rebase?

peetucket:
rebased

PR #85, opened 2015-11-11
Update to Spotlight 0.7

cbeer:


db/schema.rb
```diff
@@ -11,323 +11,326 @@
...
-ActiveRecord::Schema.define(version: 20_151_016_191_759) do
-  create_table 'bookmarks', force: :cascade do |t|
-    t.integer 'user_id', null: false
-    t.string 'user_type'
-    t.string 'document_id'
-    t.string 'title'
-    t.datetime 'created_at'
-    t.datetime 'updated_at'
-    t.string 'document_type'
+ActiveRecord::Schema.define(version: 20151111231307) do
```
ndushay: i'm guessing the spacing and quote style here isn't rubocop happy.

cbeer: I think we weren't excluding `./db/*` at first. Because this file, like most in `./db`, are automatically generated and maintained by Rails, I'd be wary (or, at least, disinterested) about enforcing our project style on them.

PR #86, opened 2015-11-12
use exhibits_solr_conf solr config files for testing for consistency

ndushay:
... and add desc to rake ci task

connects to sul-dlss/sul_exhibits_template#78


cbeer:
:+1:

PR #91, opened 2015-11-13
Trigger a reindex request for resources created by the batch PurlResources UI

cbeer:


ndushay:
The words make sense, but I still don't understand the problem, exactly.

cbeer:
This restores the previous behavior we got for free when creating a new resource automatically triggered a reindex.

This is the "model" backing this form:

screen shot 2015-11-12 at 6 46 55 pm

ndushay:
ah!  thank you!

PR #99, opened 2015-11-14
Add collapsable hit highlighting

cbeer:
A quick attempt at https://github.com/sul-dlss/spotlight/issues/1175

screen shot 2015-11-13 at 4 12 30 pm


ggeisler:
:+1: 

I can only go by the screenshot at the moment, but I wonder if the font size for the "Preview text..." toggle might be better if slightly smaller (0.75em)? It feels slightly too big of a jump from the metadata field labels to me, maybe giving it more importance than it deserves. (I'd also be fine with shipping this now and revisiting the styling details later.)

cbeer:
Yeah, the sketch in https://github.com/sul-dlss/spotlight/issues/1175 didn't give me much to go on, so I just threw the bootstrap `h4` style on it, assuming we'd do some design.

Before we go much further, though, I think it'd be useful to discuss whether this feature is useful for Feigenbaum (or other exhibits) and/or how much more work we should do.

cbeer:
@ndushay: do you know anything about the [Solr Boundary Scanners](https://cwiki.apache.org/confluence/display/solr/FastVector+Highlighter#FastVectorHighlighter-UsingBoundaryScannerswiththeFastVectorHighlighter)? I'm pretty sure it won't help us here, given the quality of the OCR, but might it be adaptable to find page-level boundaries?

ndushay:
Looks like Solr Boundary Scanners  avail are   break iterator  (WORD, LINE, SENTENCE, and CHARACTER.)  and  simple, which seems to look for specific chars.   Given OCR junk, having a particular character in our full text to indicate page breaks seems dicey.  :-(

ndushay:
So far we, have concluded that existing collections with decent quality English full text are:

- Jarndyce
- Homebrew
- GATT  (but from TEI)

And that for the future, we should determine how to accession full text created by Abbyy digitization to facilitate indexing (or how to adjust indexing to take the full-text from page level and document level pdfs)

ndushay:
See https://github.com/sul-dlss/spotlight/issues/1227  for more info on files ABBYY creates and where the full text is, etc.

PR #106, opened 2015-11-17
Update blacklight dependency; fixes #97

### LINKED ISSUE #97: Search results (gallery view) should not render a blank image tag when no thumbnail is present
Search results:

When viewing search results for objects with no image assets, there are broken thumbnails shown on search result views (e.g. https://exhibits.stanford.edu/feigenbaum/default-exhibit/browse/sound-recordings)

Initial implementation could be as simple as a blank or generic placeholder thumb if none is found; or could be more elaborate if we know the media type (audio/video/text) and could do some media specific generic icon

Detail Page:

Seems to be ok with current embedding widget.

cbeer:


cbeer:
I see a funny merge commit in here -- @peetucket did you click github's "Update branch" button here (which, I think we've discovered before, "helpfully" adds it).

peetucket:
yes, i presume because the branch was out of date - first time i used the ui to do this, i was expecting it to a rebase, but it did a merge instead.

cbeer:
Ok. Just making sure we know why. I'll mention it at standup tomorrow.

PR #109, opened 2015-11-19
Update application to support multitenancy

### LINKED ISSUE #100: Merge exhibits_requests application into this application
During this series of sprints, one common reaction has been against the number of gems + applications that make up the exhibits stack. I suggest we merge the exhibits_requests application (functionality.. not the code) and spotlight-dor-resources gem into this repository.

I believe this is an artifact of how spotlight was initially developed, coupled with a desire to keep our spotlight instance fairly close to vanilla spotlight. There may also be an element of (unrealized) flexibility in this architecture, allowing widely different exhibits to share common code.

In addition to trying to simplify the architecture, I think we'll get a couple other benefits:

- shorter release cycles for new exhibit functionality (easier to deploy; don't need to cut releases of all components; easier to run the stack locally)

- less custom components (ripping out custom capistrano deploy; can use our choice of background job processors for exhibits)

- faster exhibit creation using multi-tenancy in Spotlight instead of multi-tenant Apache

- more easily load-balanced application

- "cleaner" urls for exhibits (by removing the "default-exhibit" slug)

Downsides are:

- less flexibility to respond to exhibit-specific needs (which, again, have yet to really come up)

- more coupling between exhibits (hopefully mitigated by load-balancing the application)

- time to migrate exhibits_requests functionality into this application (hopefully just minimal, controller-level changes needed, hopefully quickly regained by faster time to deploy) 

- migration time from current architecture to a multitenant one (for one, I think we'll want to improve the import/export tooling)

This change would also let us contribute, upstream, some of our "super admin" behaviors and exhibit creation workflows.

- [x] Add exhibits index to spotlight
- [x] Add sitemap generation to spotlight
- [ ] Update sul-exhibits-template application to be multitenant
- [ ] Add requests dashboard to sul-exhibits-template
- [ ] deploy to exhibits-prod-b with shared data going to the shared NFS mount
- [ ] Migrate existing collections into multitenant application and reindex
- [ ] Add redirects from "default-exhibit" urls to new urls
- [ ] Point exhibits.stanford.edu at exhibits-prod-b
- [ ] Decommission exhibits-prod and rebuild as exhibits-prod-a
- [ ] Rebuild exhibits-prod-b and add it to the pool

### LINKED ISSUE #101: Use sul-styles gem for common styling, iconography, and CSS variables

cbeer:
Fixes #100 
Fixes #101 

## Uses the SUL footer

screen shot 2015-12-01 at 10 19 12 am

## Uses the SearchWorks color palette and fonts

screen shot 2015-12-01 at 10 20 08 am

screen shot 2015-12-01 at 10 20 50 am


ggeisler:
Looks great to me. I'm not sure we want to use the same styling as SearchWorks in the long run, but this would be a great place to start from.

I'll merge after you have a chance to rebase.

cbeer:
:ok_hand: I think this can be shipped now, as long as we're ok with using the new theme with our current exhibits.

ggeisler:
I'm okay with using the SW theme in the short run. I just think there needs to be a larger discussion about using it permanently. But given the complexity of this update, it seems better to get the new code in place, and then decide if we want to tweak the styling a bit to make sure the exhibits have a distinct identity from SW.

PR #110, opened 2015-11-19
[PROPOSAL] Raise the SCSS selector depth limit to 3

cbeer:
I think we should raise the SCSS selector depth from 2 to 3 (the scss-lint default; hound is enforcing 2..)

We're dealing both with a lot of legacy CSS, and a pretty complex CSS inheritance across multiple projects. Add to that how many HTML templates are outside our immediate control, and it seems like we get into trouble fast.

From the scss-lint docs:

```
Bad: selectors with depths of 4

.one .two .three > .four {
  ...
}

.one .two {
  .three > .four {
    ...
  }
}

Good

.one .two .three {
  ...
}

.one .two {
  .three {
    ...
  }
}
```

### DIFF
--- a/.hound.yml
+++ b/.hound.yml
@@ -1,2 +1,5 @@
 ruby:
   config_file: .rubocop.yml
+
+scss:
+  config_file: .scss-lint.yml
--- a/.scss-lint.yml
+++ b/.scss-lint.yml
@@ -1,0 +1,8 @@
+scss_files: 'app/assets/**/*.scss'
+
+exclude: 'vendor/**'
+
+linters:
+  SelectorDepth:
+    enabled: true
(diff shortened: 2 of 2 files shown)


cbeer:
This seems like something to discuss at/after standup. I chose 4 relatively arbitrarily (although my WIP #109 has some 4s), but could be fixed up to be 3, if we wanted.

ggeisler:
:+1: I'd lean towards 4 because there is some much existing SCSS, but I could certainly agree with 3. Going with 3 would likely require some refactoring time in some of the files, but that might not be a bad thing.

ndushay:
seems like this is ready to merge, yes?  Per our discussion this morning?

PR #113, opened 2015-11-23
Update spotlight dependency and use new spotlight root path

cbeer:
This should be sufficiently backwards-compatible for our purposes. All our applications only contain a single exhibit, so we'll trigger the `spotlight/exhibits#index` redirect behavior, mimicking existing behavior.


ndushay:
Are all the gem updates tied directly to the spotlight gem update?  If not, perhaps separate commits?  o.w. looks fine to me.

cbeer:
Sure, that makes sense. Those were only blacklight-spotlight dependencies, so let me `bundle update` everything.

ndushay:
uh oh - some gem update made it unhappy.  :cry:

cbeer:
Ah, I forgot about @jkeck's changes that got released in spotlight 0.8.1.

cbeer:
@ndushay :cry: => :smile:

PR #124, opened 2015-11-30
Use shared factories

cbeer:


Gemfile
```diff
@@ -52,9 +52,9 @@ group :development, :test do
 end
 
 group :test do
-  gem 'rspec-rails', require: false
-  gem 'factory_girl', require: false
-  gem 'capybara', require: false
+  gem 'rspec-rails'
+  gem 'factory_girl_rails'
+  gem 'capybara'
```
aeschylus: I may not be the most efficient reviewer, but perhaps that will be a good thing. "require:false" seems crufty, but can you briefly point me to a reason it's been changed?

Gemfile.lock
```diff
@@ -419,8 +422,8 @@ GEM
       netrc (~> 0.7)
     retriable (1.4.1)
     retries (0.0.5)
-    roar (1.0.3)
-      representable (>= 2.0.1, <= 3.0.0)
+    roar (1.0.4)
+      representable (>= 2.0.1, < 2.4.0)
```
aeschylus: Seems to be lots of version bumping; usually a good thing.

spec/factories/exhibits.rb
```diff
@@ -1,6 +0,0 @@
-FactoryGirl.define do
-  factory :exhibit, class: Spotlight::Exhibit do
-    sequence(:title) { |n| "Exhibit Title #{n}" }
-    published true
-  end
-end
```
aeschylus: Goodbye exhibits?

cbeer: The `spotlight` gem now provides this factory for us.

spec/rails_helper.rb
```diff
@@ -3,12 +3,7 @@
 require 'spec_helper'
 require File.expand_path('../../config/environment', __FILE__)
 require 'rspec/rails'
-
-# Add additional requires below this line. Rails is not loaded until this point!
-
-require 'factory_girl'
-FactoryGirl.definition_file_paths = [File.expand_path('../factories', __FILE__)]
-FactoryGirl.find_definitions
+require 'factory_girl_rails'
```
aeschylus: Seems like a good replacement that doesn't require a certain ordering.

Gemfile
cbeer: Unfortunately, I don't know why it was done this way in the first place.

aeschylus: Fair enough.

Gemfile.lock
cbeer: Yep. I took a look at this too.. They're doing a major refactor in `3.0`, and `2.4` introduces some deprecation warnings that are internal to `roar`. They're being good citizens by not inflicting on us deprecation warnings that we're helpless to do anything about.

spec/rails_helper.rb
cbeer: Yep, `factory_girl_rails` handles all of this for us. Not sure why we weren't using it before..

Gemfile
ndushay: possibly because if someone is doing a bundle install mistakenly with the test group when they don't need it, it's okay?   The whole "require: false" thing is documentation I've encountered elsewhere for rspec or coveralls or something like that and I suspect it's just a pattern the coder was used to.

cbeer: Perhaps. RSpec isn't pushing it, so it seems safe to skip.

PR #132, opened 2015-12-03
Override bootstrap navbar colors to use our beige palette

cbeer:
After:

screen shot 2015-12-02 at 4 40 55 pm


ndushay:
Does @ggeisler  need to review this?

PR #144, opened 2015-12-10
Integrate devise inevitable with our WebAuth implementation

jkeck:


ggeisler:
@jkeck Can you post or send me an example of the generated email text? (Just so I don't have to pull the branch or piece together how it reads with the locale strings + variables.)

jkeck:
@ggeisler yeah here it is, and @cbeer noticed a typo in there (although I think we'll need to fix that in Spotlight too).

```
To: User Name <user-name@stanford.edu>
Subject: Invitation instructions

Hello user-name@stanford.edu!

The Exhibits Administrator has invited to be a member of 'Invite Test' at http://exhibits-stage.stanford.edu/default-exhibit/home. You can accept it through the link below.

[link]Accept invitation[/link]

If you don't want to accept the invitation, please ignore this email.
Your account won't be created until you access the link above.
```

ggeisler:
Thanks. This might be what @cbeer noticed but the first sentence needs a "you":

>
The Exhibits Administrator has invited you to be a member of 'Invite Test' at http://exhibits-stage.stanford.edu/default-exhibit/home.

But how about we also make a slight update to it and the sentence that follows it so they read like this:

>
The SUL Exhibits Administrator has invited you to help work on the "Invite Test" exhibit at http://exhibits-stage.stanford.edu/default-exhibit/home. You can accept this invitation by clicking the link below.

Also suggest a slight rewording of the last part, just so people don't have the potential confusion of thinking we're going to create some other, new account they have to worry about (because in our instance, we're not ever creating a new account, correct? Just activating their SUNet with a role in the exhibit?):

>
If you don't want to accept the invitation, please ignore this email. You won't be added to the "Invite Test" exhibit team until you access the link above.

(Note I changed the quotes from single to double, since for non-programmers I think double-quotes are more common and familiar - well, in the U.S. at least.)

ggeisler:
Sorry, forgot to address the subject line. How about we update that to something like this, just to make it feel a bit less generic:

>

Subject: SUL exhibit invitation instructions

jkeck:
Sounds good @ggeisler, making those changes now.

Just to confirm, only adding the `you` should be done in Spotlight itself and the rest of the changes are local customizations?

jkeck:
Okay, I think we're ready to :shipit:.  Let me know if there are any other changes we need to make.

ggeisler:
:+1: Thanks!

PR #148, opened 2015-12-11
Adjust font sizes for various headings

cbeer:


ggeisler:
Can you provide screenshots so I can see what the changes look like?

cbeer:
I was afraid somebody would ask for that. Guess I need to get some collections going locally..

ggeisler:
Well, I'm sure the changes are good. We could merge them and if I think something needs adjusting we can deal with that in another commit.

cbeer:
:+1: I think these were the major changes we discussed, and there will probably be others.

PR #172, opened 2016-01-12
update exhibits cap prod deployment to automatically deploy to both -…

peetucket:
…a and -b servers automatically

### DIFF
--- a/config/deploy/production.rb
+++ b/config/deploy/production.rb
@@ -1,7 +1,12 @@
-set :deploy_host, ENV['CAPISTRANO_PRODUCTION_DEPLOY_HOST'] || ask(:deploy_host, '')
+set :deploy_host, ENV['CAPISTRANO_PRODUCTION_DEPLOY_HOST'] || ask('deploy_host prefix (leave off stanford.edu)', '')
 set :user, ENV['CAPISTRANO_PRODUCTION_USER'] || ask(:user, '')
+set :bundle_without, %w(deployment development test).join(' ')
 
-server "#{fetch(:deploy_host)}", user: fetch(:user), roles: %w(web db app)
+%w(a b).each do |suffix|
+  server "#{fetch(:deploy_host)}-#{suffix}.stanford.edu",
+         user: fetch(:user),
+         roles: %w(web db app)
+end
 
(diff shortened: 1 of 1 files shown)


config/deploy/production.rb
```diff
@@ -1,9 +1,15 @@
-set :deploy_host, ENV['CAPISTRANO_PRODUCTION_DEPLOY_HOST'] || ask(:deploy_host, '')
+set :deploy_host, ENV['CAPISTRANO_PRODUCTION_DEPLOY_HOST'] || ask('deploy_host (leave off stanford.edu)', '')
```
ndushay: perhaps the prompt should be clear that this is the prod box prefix, not the full prod box name?

config/deploy/production.rb
```diff
@@ -1,9 +1,15 @@
...
 set :user, ENV['CAPISTRANO_PRODUCTION_USER'] || ask(:user, '')
+set :bundle_without, %w(deployment development test).join(' ')
```
ndushay: i think this was happening automagically?  Not sure why - puppet config for bundler?

config/deploy/production.rb
```diff
@@ -1,9 +1,15 @@
...
 
-server "#{fetch(:deploy_host)}", user: fetch(:user), roles: %w(web db app)
+user_server_extensions = %w(a b)
+user_server_extensions.each do |extension|
```
ndushay: %(a b).each do  |suffix|  ... ?

ndushay:
My comments are quibbly, except maybe the one about being clear we are entering the machine name prefix, not the full host name?

config/deploy/production.rb
peetucket: if so, that would be cool, but i think this just determines how bundler is called by capistrano when deploying - we use it in other projects i deploy with (and SW does too)

when he has a chance @cbeer can let us know

config/deploy/production.rb
peetucket: good idea, i'll adjust

ndushay:
:+1:    (rebase and merge yourself)

cbeer:
Fixed by #186

PR #192, opened 2016-03-02
use latest version of spotlight-dor-resources gem

peetucket:


Gemfile
```diff
@@ -75,7 +75,7 @@ gem 'blacklight', '~> 5.16'
 gem 'blacklight-gallery', '~> 0.3'
 gem 'blacklight-maps', '0.2.0'
 gem 'blacklight-spotlight', '~> 0.16'
-gem 'spotlight-dor-resources', '~> 1.0'
+gem 'spotlight-dor-resources', '~> 1.0.1'
```
cbeer: Could we say `'~> 1.0', '>= 1.0.1'`?

peetucket: Yup we can do that, but I wasn't sure if there was a precedent to pin this way to prevent a major version upgrade from breaking things unintentionally (with a bundle update).

peetucket: whoops, never mind, you're code will do that

Issue #193, opened 2016-03-03
Inviting new admin user doesn't work on -stage


ggeisler:
We want to give a (Stanford) person access to an exhibit on exhibits-stage but the 'send an invite to new use'r feature doesn't see to work on -stage:

![configuration_-_site_configuration_-_users___martine_madill_-_online_exhibits__stage_](https://cloud.githubusercontent.com/assets/101482/13505044/05ef62c2-e12c-11e5-9b75-993b9544b7b4.png)

ggeisler:
The problem also appears to exist on -prod

PR #194, opened 2016-03-03
Update invitation instructions to match upstream changes

cbeer:


ggeisler:
Thanks for fixing this so quickly.

ggeisler:
Not sure this fix worked. I deployed to -stage but still see the same problem. 

I noticed that when clicking the Save changes button after entering the new user email address there was a brief display of the intended "... send an invite?" message, but it only showed for a split second and then the page updaed to look like the screenshot in #193.

PR #195, opened 2016-03-14
Set action_mailer's default options from the environment-specific settings

cbeer:


config/initializers/action_mailer.rb
```diff
@@ -1,3 +1,9 @@
-if defined? Settings and Settings.action_mailer and Settings.action_mailer.default_url_options
-  Exhibits::Application.config.action_mailer.default_url_options = Settings.action_mailer.default_url_options.try(:to_h) || {}
+if defined? Settings and Settings.action_mailer
```
jkeck: Hate to nitpick (particularly since you're just copying this from above) but can we turn this `and` into an `&&`?

Other than that, LGTM!

cbeer: :ok_hand:

PR #236, opened 2016-07-08
Restrict thumbnail images to no more than 400px wide

### LINKED ISSUE #207: Reevaluate how we present item row images
If you use the item row widget with a single item, and that item is a landscape image, it can be stretched wider than than its natural size (because the image is requested at 400px max and the CSS sets its height to 250px, very wide images need more than 400px width to fit the 250px height).

Example: https://exhibits.stanford.edu/david-rumsey-map-collection/feature/views-of-the-west

We need to look at the issue more closely to decide whether one or more of the below would handle the issue:

- more versatile CSS for the Item Row widget (using `max-height` instead of `height` maybe?)
- special-case CSS for when the Item Row widget only has one item
- requesting a larger (e.g., 800px max) image when the Item Row widget only has one item

cbeer:
To fix #207, we could limit all thumbnails to no more 400px wide.

screen shot 2016-07-08 at 4 38 40 pm

cc @ggeisler

### DIFF
--- a/app/assets/stylesheets/modules/spotlight_overrides.scss
+++ b/app/assets/stylesheets/modules/spotlight_overrides.scss
@@ -183,4 +183,5 @@
 
 .thumbnail > img {
   @include thumbnail-image-border;
+  max-width: 400px;
 }


ggeisler:
Sure, that's a straightforward way to handle it. Especially if we are always requesting thumbnail images that are 400px max, it makes sense not to allow them to display larger than that. I was thinking something like the panorama image might look better if we could find a solution to enable it to display wider, but it looks fine at 400px, so I like this solution.

PR #239, opened 2016-07-12
Hide exhibit creation from non-admin users. Fixes #212

### LINKED ISSUE #212: Restrict visibility of Request new exhibit button to super-admins
We currently show the `Request new exhibit` button to any user viewing the Exhibits homepage. For the near-term timeframe, however, we want curators to follow a formal request process rather than create new exhibits themselves. So we should restrict visibility of the `Request new exhibit` button to only super-admin users (if we create an additional class of users to include non-super-admin members of the Exhibits service team, we might want to also enable that group to see the button and create new exhibits).

online_exhibits

If the `Create new exhibit` link in the user pulldown menu appears for non-super-admin users (I'm not sure how things look for WebAuthed non-super-admin users), we should restrict visibility of it in the same way.

test_exhibit_one_-_online_exhibits

mjgiarlo:
Use the existing `:create, Spotlight::Exhibit` ability to determine who sees exhibit creation links.

Fix a typo in the `shared/_site_sidebar.html.erb` partial.

Override the Spotlight `add_default_roles` method from our `User` model; otherwise all user instances wind up being administrators in the test env. We do not use this built-in functionality, instead relying on Stanford WebAuth group membership to determine who gets administrative access. **N.B.**: this was required for the prior feature spec-based tests to pass, and is not required for this PR. I left it in so it doesn't trip us up in the future and added a model spec to cover the change. If you'd rather see this dropped from this PR, or broken into separate commits, let me know. Otherwise, :shipit:.

### DIFF
--- a/app/views/shared/_site_sidebar.html.erb
+++ b/app/views/shared/_site_sidebar.html.erb
@@ -19,9 +19,11 @@
     </li>
   </ul>
 
-  <%= link_to spotlight.new_exhibit_path, class: "btn btn-default", role: "button" do %>
-    Request a new exhibit
-    <small class="sunet-notice">Requires SUNet login</small>
+  <% if can?(:create, Spotlight::Exhibit) %>
+      <%= link_to spotlight.new_exhibit_path, class: "btn btn-default", role: "button" do %>
+          Request a new exhibit
+          <small class="sunet-notice">Requires SUNet login</small>
+      <% end %>
   <% end %>
 
   <% if can?(:manage, Spotlight::Site.instance) || can?(:create, Spotlight::Exhibit) %>
@@ -33,7 +35,7 @@
         <li><%= link_to t(:'spotlight.sites.edit.page_title'), spotlight.edit_site_path %></li>
       <% end %>
 
-      <% if can? :manange, Spotlight::Exhibit %>
+      <% if can? :manage, Spotlight::Exhibit %>
         <li><%= link_to t(:'spotlight.sites.edit_exhibits.page_title'), spotlight.edit_site_exhibits_path %></li>
         <li><%= link_to t(:'spotlight.admin_users.index.page_title'), spotlight.admin_users_path %></li>
       <% end %>
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -31,6 +31,12 @@ class User < ActiveRecord::Base
     end
   end
 
+  # Exhibits will populate its own roles from workgroups, and doesn't need
+  # this upstream feature.
+  def add_default_roles
+    # no-op
+  end
+
   private
 
   # rubocop:disable Metrics/MethodLength
--- a/spec/models/user_spec.rb
+++ b/spec/models/user_spec.rb
@@ -14,4 +14,9 @@ describe User do
       expect(subject).to be_a_superadmin
     end
   end
+  describe '#add_default_roles' do
+    it 'does not inject the administrative role' do
+      expect(subject).not_to be_a_superadmin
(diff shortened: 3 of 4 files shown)


app/models/ability.rb
```diff
@@ -12,6 +12,7 @@ def initialize(user)
     can :manage, Spotlight::Resources::DorHarvester, exhibit_id: user.exhibit_roles.pluck(:resource_id)
 
     # disable spotlight functionality we don't want to expose in spotlight:
+    can :create, Spotlight::Exhibit if user.superadmin?
```
mjgiarlo: I don't believe this line is necessary -- this ability already exists. Part of the testing I was doing, so feel free to ignore in terms of overall code quality for now.

app/views/shared/_site_sidebar.html.erb
```diff
@@ -33,7 +35,7 @@
         <li><%= link_to t(:'spotlight.sites.edit.page_title'), spotlight.edit_site_path %></li>
       <% end %>
 
-      <% if can? :manange, Spotlight::Exhibit %>
+      <% if can? :manage, Spotlight::Exhibit %>
```
mjgiarlo: Unrelated typo fix.

spec/features/exhibit_index_spec.rb
```diff
@@ -1,7 +1,42 @@
...
+  before do
+    login_as(user, scope: :user)
     visit '/'
   end
+
+  describe 'user pull-down menu' do
+    context 'with a non-admin user' do
+      let(:user) { FactoryGirl.create(:curator) }
+      it 'hides "create exhibit" link' do
+        # This fails because user is an admin (inexplicably)
```
mjgiarlo: When I drop in through `byebug` and call `user.superadmin?`, I get `true`. I get the same with a totally vanilla FG user as well.

spec/features/exhibit_index_spec.rb
```diff
@@ -1,7 +1,42 @@
...
+        expect(page).to have_link('Create Exhibit')
+      end
+    end
+  end
+
+  describe 'new exhibit block' do
+    context 'with a non-admin user' do
+      let(:user) { FactoryGirl.create(:curator) }
+      it 'hides "request new exhibit" link' do
+        # This fails because user is an admin (inexplicably)
```
mjgiarlo: Same as above.

spec/support/devise.rb
```diff
@@ -0,0 +1,4 @@
+require 'warden'
```
mjgiarlo: Need this for `login_as` in feature specs.

app/views/shared/_site_sidebar.html.erb
```diff
@@ -19,9 +19,11 @@
     </li>
   </ul>
 
-  <%= link_to spotlight.new_exhibit_path, class: "btn btn-default", role: "button" do %>
-    Request a new exhibit
-    <small class="sunet-notice">Requires SUNet login</small>
+  <% if can?(:create, Spotlight::Exhibit) %>
```
jcoyne: Is creating the same as requesting?  I thought they were different.

cbeer: Nope, same thing.

jcoyne: 🆗

cbeer:
Spotlight grants the first user the superadmin role:

https://github.com/projectblacklight/spotlight/blob/master/app/models/concerns/spotlight/user.rb#L26

PR #242, opened 2016-07-12
Tweak rubocop behavior

mjgiarlo:
* Display cop names to more easily troubleshoot Rubocop offenses
* Remove exclusions that do not cause Rubocop offenses
* Remove `RSpec/DescribeClass` exclusion in favor of marking feature specs with `type: :feature`
* Remove `RSpec/FilePath` exclusion (has no effect on # of offenses)
* Require rubocop-rspec in TODO file to silence warning
* Push `rubocop` task down as dependency of `ci` task, and fail early if offenses are found
* Fix offense in `spec_helper.rb`

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -2,13 +2,15 @@
 # for example lib/tasks/capistrano.rake, and they will automatically be available to Rake.
 require File.expand_path('../config/application', __FILE__)
 
-task default: [:ci, :rubocop]
+task default: :ci
 
 Exhibits::Application.load_tasks
 
 begin
   require 'rubocop/rake_task'
-  RuboCop::RakeTask.new(:rubocop)
+  RuboCop::RakeTask.new(:rubocop) do |task|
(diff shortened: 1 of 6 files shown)


Rakefile
```diff
@@ -19,7 +21,7 @@ rescue LoadError
 end
 
 desc 'Run tests in generated test Rails app with generated Solr instance running'
-task ci: [:environment] do
+task ci: [:rubocop, :environment] do
```
jcoyne: I really like having a task to run that spins up solr & runs tests without doing rubocop. But I can fix that in another PR.

mjgiarlo: OK. I assume it's because you prefer to defer Rubocop-related changes until you're sure everything is :green_apple: and :smiley:?

jcoyne: Yeah, I want my tests to run as fast as possible when I'm doing TDD

PR #244, opened 2016-07-12
Update dependencies

mjgiarlo:
Updates all outdated dependencies highlighted by Gemnasium except for the following:

* rails (cannot update to Rails 5 yet because Spotlight can't yet use it)
* devise-remote-user (this is a false positive: devise-remote-user is already at its latest version, 0.4.1. Gemnasium falsely reports that 0.5.0 is released, when in fact it was yanked)
* web-console (3.3.1 seems to want to move our `activemodel` version)


cbeer:
I don't think we should upgrade blacklight-maps yet. I don't think we have indexed the data in a way that'll work with the new geojson-based whatnot.

Gemfile
```diff
@@ -59,7 +58,7 @@ group :test do
 end
 
 group :production do
-  gem 'mysql2', '~> 0.3.20'
+  gem 'mysql2', '~> 0.4.4'
```
cbeer: Does this actually work now? We had to pin before because activerecord 4.x wasn't happy with it.

cbeer: Looks like maybe: https://github.com/rails/rails/commit/2dfff4d31634d669f4c198c4fb85b59d39477862

Gemfile
```diff
@@ -73,7 +72,7 @@ gem 'bootstrap-sass', '~> 3.3.5'
 
 gem 'blacklight', '~> 6.3'
 gem 'blacklight-gallery', '~> 0.3'
-gem 'blacklight-maps', '0.2.0'
+gem 'blacklight-maps', '0.5.0'
```
cbeer: I don't think we should upgrade blacklight-maps yet. I don't think we have indexed the data in a way that'll work with the new geojson-based whatnot.

Gemfile
```diff
@@ -74,7 +73,7 @@ gem 'bootstrap-sass', '~> 3.3.5'
 gem 'blacklight', '~> 6.3'
 gem 'blacklight-gallery', '~> 0.3'
 gem 'blacklight-maps', '0.2.0'
-gem 'blacklight-spotlight', '~> 0.18'
+gem 'blacklight-spotlight', '~> 0.19'
```
cbeer: I think you need to run `rake spotlight:install:migrations`?

cbeer: (and here's a handy convenience task for next time: https://github.com/sul-dlss/exhibits/blob/master/lib/tasks/spotlight.rake#L46)

mjgiarlo: Good catch.

cbeer: Burned me enough times.

PR #279, opened 2016-07-28
[WIP] Use the iiif_crop branch of spotlight

jcoyne:
This branch exists so we can deploy to staging and review the work so far.


ggeisler:
I like the responsiveness (in terms of speed) of the whole cropping interaction if you just move or adjust the crop box and Save. However, it does seem that the zooming aspect of this new approach brings up some issues with the interaction we should discuss:

- The actual manipulation of the crop box (moving or resizing it) seems fine to me. My concerns are just related to the zooming aspect. I feel like I did get some unexpected jumpiness with the crop box a couple of times, but I can't reproduce it and it didn't happen enough that I was bothered by it.

- A big concern I do have is the lack of visual affordance for zooming. Because clicking outside the crop box zooms in, it will be pretty easy for a user to inadvertently zoom in on the image, and there is no visual indication of how to zoom back out, and no UI control for zooming. This is my biggest concern because our exhibit curators are generally not very advanced technical users, and without UI controls (or even keyboard controls?) for zooming I suspect some will be frustrated and end up having to figure out to refresh the page to start over (some might figure out how to zoom back out, but I think a lot won't).

- It's pretty easy to get into a situation where you lose the crop box. For example, I just clicked in the image twice (outside of the crop box) and I can no longer see the crop box. That might confuse unsophisticated or new users (and again, because they might not know how to zoom out, they'd have to refresh the page to get out of the situation).

I don't understand the details, but I understand that there is a (good?) reason the zooming exists in this implementation. But I do think it is the zooming that is potentially problematic here, and for Spotlight purposes I don't think the curator really ever needs to zoom in on an image to make a satisfactory crop.

This might be oversimplified but maybe options to discuss are:

1. Figure out how to prevent the image from being zoomable; just display the image within a fixed area with max width and height.

2. Stick with the current implementation but add (in order of importance, last couple not as important):
 - Visual zoom-in and -out controls
 - Keyboard controls for zoom-in and -out
 - Re-center the crop box when it goes outside the display area (this might need more thought, and might not be necessary or the best thing to do)
 - Restrict the crop box to the boundary of the image. (Currently you can save a crop where some of the crop box is off the edge of the image; I got unexpected results when the box was off the left side of the image. Though you could argue the curator should be responsible for keeping the box within the image.)

aeschylus:
Talking with Cathy yesterday and seeing the demo earlier, what is outstanding? It should be impossible for the crop region to go outside of the viewport area, and I can't reproduce the behaviour described above. The cropbox remains the same size and shape and position with respect to the containing element regardless of where the user zooms (it sticks to the viewport and it is impossible for it to leave the viewport). The Image can be locked so the user can't zoom at all (solving 1. above), which is great for most cases, and I believe this is the current configuration. 

The use case for zooming arose during the demo this morning. There was a problem cutting a long, thin strip out of the tall, thin image because the image was locked at a particular zoom level where the remaining horizontal space could not be used to adjust the cropbox more precisely. With zooming enabled, the user could zoom in until the left and right sides of the image were filling the frame, allowing the cropbox to be enlarged and tweaked more easily. With the current settings, OpenSeadragon is very very sensitive, so those would need to be tweaked to make that a more pleasant experience. There are keyboard controls built into openseadragon, but they only work if the OSD canvas has DOM focus. Focus can be applied with javascript, or the controls can be proxied by something like [keymaster](https://github.com/madrobby/keymaster), which could also manipulate the image with API calls rather than proxying the existing events.

A minimap and +/- controls could be added to improve the visual affordances.

As for the unexpected jumpiness, is this with OpenSeadragon or the cropbox? I know @jcoyne resolved an issue causing the cropbox to "jump" to center itself on the mouse if the mousedown event occurred too far from the centre of the box, but that has been resolved. As for OSD, I think almost everyone who uses it finds it too jumpy, and you have to tweak the configuration a little bit to produce the right characteristics. Some balance between sticky and "slippy" (as it keeps being described). The characteristics come from a [spring-based animation system](https://medium.com/@flyosity/your-spring-animations-are-bad-and-it-s-probably-apple-s-fault-784932e51733#.yjqtgglao). There are [configuration options](https://medium.com/@flyosity/your-spring-animations-are-bad-and-it-s-probably-apple-s-fault-784932e51733#.yjqtgglao) for spring stiffness and animation time upon initialisation. By changing these, you can converge on a set of values that feels right for the use case. Is this what is meant by "jumpiness", and would this be enough to resolve it?

After talking with @ggeisler and @jcoyne, I think the only outstanding piece is adding a check and restriction on box resizing and placement that would prevent the box from being drawn outside of the image boundary, since the behaviour that zooms the crop box is not reproducible and it is known how to turn zooming off if desired.

jkeck:
Closing this as the migrations and schema changes in this branch have changed from what actually went into master.

PR #290, opened 2016-08-01
Remove exhibits_solr_config; fixes #258

### LINKED ISSUE #258: Revisit exhibits_solr_conf gem
See: https://github.com/sul-dlss/exhibits_solr_conf/

Now that solr_wrapper has evolved, I doubt we need a separate gem just for versioning the (test-only?) schema.  If we're going outside this repo for config, we should be fetching the a specific HASH/tagged-version schema from `sul-solr-configs` so that there is only one authoritative version to worry about.  

The application needs to establish sync with a schema version, and using a specific HASH/tag accomplishes that.  Unless there are other consumers of that gem, I propose we reabsorb it and remove it.

cbeer:
Fixes https://github.com/sul-dlss/exhibits_solr_conf/issues/13
Fixes #258

### DIFF
--- a/app/controllers/catalog_controller.rb
+++ b/app/controllers/catalog_controller.rb
@@ -57,8 +57,8 @@ class CatalogController < ApplicationController
     config.view.gallery.partials = [:index_header, :index]
     config.view.masonry.partials = [:index]
     config.view.slideshow.partials = [:index]
-    config.view.maps.type = 'placename_coord'
-    config.view.maps.placename_coord_field = 'placename_coords_ssim'
+    config.view.maps.type = 'bbox'
+    config.view.maps.bbox_field = 'point_srpt'
     config.view.embed.partials = [:osd_or_embed]
     config.view.embed.if = false
 
--- a/.solr_wrapper.yml
+++ b/.solr_wrapper.yml
@@ -1,5 +1,5 @@
 # Place any default configuration for solr_wrapper here
 # port: 8983
 collection:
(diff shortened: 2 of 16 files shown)


cbeer:
@atz any ideas about the build failure here? Seems like something is unhappy with the new solr configs?

atz:
`Error: 'undefined field point_bbox__minX','code'=>400}}` at https://travis-ci.org/sul-dlss/exhibits/builds/149031804#L400

Something is still referencing point_bbox that shouldn't be.  I'll try to add some cleanup.

app/controllers/catalog_controller.rb
```diff
@@ -57,8 +57,8 @@ class CatalogController < ApplicationController
     config.view.gallery.partials = [:index_header, :index]
     config.view.masonry.partials = [:index]
     config.view.slideshow.partials = [:index]
-    config.view.maps.type = 'placename_coord'
-    config.view.maps.placename_coord_field = 'placename_coords_ssim'
+    config.view.maps.type = 'bbox'
```
atz: I don't understand what this configuration line does or why it would necessarily change here.  Can you enlighten me?

cbeer: I couldn't find any indication we've ever indexed anything into `placename_coords_ssim`, so using the srpt field which we ostensibly index into seemed like an improvement.

atz:
Not getting the failure locally during `bundle exec rake`, but it is apparently running the specs multiple times.  Oddly.

PR #308, opened 2016-08-04
Fix parameterized heatmap document path; fixes #306

### LINKED ISSUE #306: heatmap show view fails
When you click on one of the results in the map, it goes to this url: `https://exhibits-stage.stanford.edu/map-view-testing/catalog/id?exhibit=map-view-testing`, this page fails with a 404

cbeer:


cbeer:
We've previously talked about other ways to handle populating the map detail view; this hack works for now, but might encourage us to look at better ways (like, providing a rendered partial for each document within the json search results format? Or, even incrementally better, like add the document URL to the search results json)

PR #333, opened 2016-08-18
Clean up 2 tests

atz:
Use common let statements.
Put let statements together.

### DIFF
--- a/.hound.yml
+++ b/.hound.yml
@@ -1,4 +1,5 @@
 ruby:
+  enabled: false
   config_file: .rubocop.yml
 
 scss:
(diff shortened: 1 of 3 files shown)


atz:
What is the point of having Hound pass the build but then Travis fail it by running rubocop?  Shouldn't that be the same as Hound failing the build?

Also, don't care about line length on those.  I prefer the long-line versions on those.

mjgiarlo:
@atz Fwiw, this is why i turned off Ruby checking via Hound in Sufia: https://github.com/projecthydra/sufia/blob/master/.hound.yml I'd support doing the same here

atz:
@mjgiarlo rebased and tweaked, including hound as you recommended.

PR #392, opened 2016-12-07
Move fetching collection information into a background job

### LINKED ISSUE #391: Ability to add new items to an exhibit has stopped working
Currently, as a curator, when I attempt to add a new druid or druids to the item list on the admin console of two different published exhibits (Hanna House & Stanford Oral History), the result I see is an
screen shot 2016-12-07 at 9 53 25 am
 error message (attached) "The exhibits.stanford.edu page isn't working. exhibits.stanford.edu is currently unable to handle this request" Subsequent to seeing this error message, I have searched for the items I attempted to add, in case the error message was incorrect -- and I have verified the items I attempted to add were not added, as they did not appear as a search result.

As a curator, I expect to be able to add new druids to a published or unpublished exhibit where I am an administrator. After I add a new druid or druids to the item list in the admin console, I expect to see confirmation when my items have been successfully added. I also expect to be able to search for the newly added items in my exhibit and have them appear as a search result.

cbeer:
This might fix #391, where exhibits with excessively long lists of druids where we hit an HTTP timeout collecting this information.

### DIFF
--- a/app/jobs/record_resource_metadata_job.rb
+++ b/app/jobs/record_resource_metadata_job.rb
@@ -1,0 +1,6 @@
+# Record the metadata for an index job
+class RecordResourceMetadataJob < ActiveJob::Base
+  def perform(resource)
+    resource.update_collection_metadata!
+  end
+end
--- a/app/models/dor_harvester.rb
+++ b/app/models/dor_harvester.rb
@@ -22,7 +22,8 @@ class DorHarvester < Spotlight::Resource
 
   def waiting!
     super
-    update(collections: fetch_collection_metadata)
+    update(collections: {})
+    RecordResourceMetadataJob.perform_later(self)
   end
 
   def collections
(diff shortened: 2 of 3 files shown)


jkeck:
This LGTM.  Just one question: Is it [this code](https://github.com/sul-dlss/exhibits/pull/392/files?diff=unified#diff-9a7ba40e3d71dca813d5ccc4d3099a3cR81) that is potentially slow?  Does that hit Dor somehow?  Just trying to figure out what is generating wait time of > 60 seconds.

cbeer:
Yes, definitely slow. It's hitting an external service for each druid. By moving it into a background job, we'll avoid the rack timeout.

jkeck:
Gotcha.  And is that in the `collection?` method or somewhere else in building the `resources` array? (just curious).

cbeer:
`exists?`, `collection?`, and `#items` all hit external services.

Issue #388, opened 2016-12-02
Implement advanced search as an option for an exhibit


peetucket:
As an exhibits user, I need to have greater control over search so I can find very specific items.

Use https://searchworks.stanford.edu/advanced as an example.  A user in an exhibit where advanced search has been enabled should be able to visit a separate "advanced search" page where they are shown a series of pre-configured fields from the exhibit with search boxes next to each, a drop-down menu to select "any or all", a drop menu to select sort preference, and several standard facets to filter by.  They should be able to use standard boolean operators in each text field (e.g. AND, OR and NOT) as well as fuzzy search operators (e.g. ? and *).  They need to only fill in a single text field, but can fill in as many as they want.  Clicking the "search" button executes the search and returns the standard exhibits search results page.

Acceptance criteria:
- [ ] a new configuration option per exhibit is available allowing advanced search to be enabled.  Suggest a checkbox in "Dashboard --> Configuration --> Search --> Options"
- [ ] an advanced search page is available and linked to from the navigation if configuration above selected
- [ ] a list of fields with corresponding search text boxes is provided on the advanced search page  
- [ ] the fields to be shown on the advanced search page are configurable in the same way the current "field-based search" is configured (in "Dashboard --> Configuration --> Search --> Options").  A superset of fields that is available to select from to be determined later but starting with Title, Author/Contributor, Subject, Call number, Full text, Year.
- [ ] boolean operators are supported in each text box (AND, OR and NOT)
- [ ] fuzzy search operators are supported in each text box (? and *)
- [ ] a drop down menu for specifying sort order is available and shows the same fields currently available for sorting on the standard search results page
- [ ] a list of facets to filter by is available, and shows the facets currently selected in the Exhibit's "Dashboard --> Configuration --> Search --> Facets" panel
- [ ] a search tips panel is provided exactly as shown on https://searchworks.stanford.edu/advanced

ggeisler:
At least some of your acceptance  criteria feel an awful lot like specific **design** decisions. For example, is saying there should be a search tips panel "exactly like this" really appropriate as an acceptance criterion? 

Also, maybe you've already discussed this with @jkeck and/or @cbeer, but is this really an Exhibits ticket  (versus Spotlight repo)?

snydman:
For Parker on the Web there is an advanced search page at https://parker.stanford.edu/parker/actions/page.do?forward=manuscript_description_search.  We need similar capability in Parker on the Web 2.0 which will be in Spotlight.  

[...]

PR #396, opened 2016-12-09
Install blacklight_advanced_search into the application

### LINKED ISSUE #322: Boolean searching (e.g. OR) not supported anymore
reported to @peetucket by Scott Van Duyne

Hi Peter,
just found this one going over the search.  On the browse page of the Feigenbaum site(https://exhibits.stanford.edu/feigenbaum/browse), the category called "Expert Systems: by the Heuristic Programming Project" returns "0 items" at the moment.  It used to return a couple hundred.

It uses a Title search like this:
Title = "CRYSALIS PUFF OR MYCIN OR HASP"

So I checked the search out live in the search box, and it really pulls up 0 items now, I also tried removing the "OR's", and that also resulted in 0 records, and I then tried selecting from Everything (instead of Title only), that resulted in 2 record returns, which were both long reports that had all four of those terms in them.  

So, the bug: apparently, when you put multiple terms in the search text box, they are all implicitly AND'ed together.  The expected behavior, till now, has been that they should be OR'd.   Also, "AND" "OR", "NOT" keywords and parens "()" used to work as expected (we even documented that Boolean expression feature in one of the "About" pages-- we would like them to keep working) but they do not seem to work anymore.  

Seems like the default config for that search box just got changed up at some point.

cbeer:
All this first pass does is adds blacklight_advances_search's custom query parsing. From the documentation:

> But queries with special operators will have appropriate Solr queries generated for them, usually including nested "query" elements, to have the same meaning they would in advanced search.

This logic is only triggered by the presence of the keywords "AND", "OR" or "NOT" (case-sensitive), so it seems unlikely to unintentionally change behavior for users.

After deploying this change, we should determine if this fixes #322 sufficiently, and determine how to reword #388.


jkeck:
Is there any useful test we could add that would indicate that we had hooked up advanced search properly?  Not sure if it would be to execute a simple search to indicate that a boolean operator worked.

I don't think we should hold up this PR because of that though, given that the nature of this is to implemented (theoretically) already tested behavior.

cbeer:
There are some dependencies we're missing to write a decent test, so I've recorded that here: https://github.com/sul-dlss/exhibits/issues/400

jkeck:
I believe the coverage reduction is not in relevant lines of code.  I believe you've added the relevant tests for the changes you've made.

PR #406, opened 2016-12-21
Update sidebar links

### LINKED ISSUE #405: Update links provided on Exhibits landing page
Reasons for update:

* We want to provide an easier way for the Stanford community to understand how to request an exhibit
* We want to provide visibility and access to the small set of documentation the Exhibits service team has recently produced

### Mockup

Suggested arrangement of the sidebar on exhibits.stanford.edu after making the updates explained below:

exhibits-landing-page-links_360

* New link "Exhibit documentation" goes to an index page in Consul, which links to the Thinking about an exhibit, Quick start guide, and Pre-Publication checklist documents. Show the Stanford "S" if this documentation requires WebAuth.

* New link "Request an exhibit" goes to a page in Consul that tells the user how to request an exhibit and mentions the caveats for requests (URL slug, are items in SDR, etc.). Show the Stanford "S" if this link requires WebAuth. (Longer term we might consider making this link go directly to a form--with the same info and caveats--so the user doesn't have to send an email to request an exhibit.)

* Remove the current button that superadmins see labeled "Request a new exhibit"

* Add a link under the Actions heading labeled "Create new exhibit." Only superadmins should see this. The link goes to the new exhibit form, previously linked to by the "Request a new exhibit" button mentioned above.

Note that the Consul pages are WIP by @caaster.

tingulfsen:
(eventually) fixes #405


app/assets/stylesheets/modules/home.scss
```diff
@@ -43,3 +43,12 @@
 #all_users.well {
   background-color: $white;
 }
+
+.nav > li > a.exhibit-inline {
```
tingulfsen: I'm not sure how to write this CSS better. The goal is for the Stanford "S" to be displayed on the same line as the link itself.

jkeck: @tingulfsen just FYI what they meant was they wanted `.exhibit-inline` instead of `a.exhibit-inline` (i.e. drop the qualifying `a` element).  I ended up tweaking this a bit to go w/ another approach so that the icon was part of the image, but just thought I would mention it in case you see this crop up again.

jkeck:
I have pushed up a commit that tweaks things a bit.  The main point of my change is to get the Stanford S icon to be part of the click target.

exhibits-sidebar

The current implementation turned the links into `inline-block` which also meant they no longer took up the entire width of the `li` (thus reducing the click-target further).

I had attempted a background image first, but that's when I came across the fact that the nav link is intended to be block element (which makes icon positioning a PITA).

Feel free to just squash my commit into @tingulfsen's if this looks good.

PR #407, opened 2017-01-04
Advanced search queries should use edismax query parser

cbeer:


jkeck:
See #779 for lots of testing/conversation/context if we pick this back up.

Issue #421, opened 2017-01-21
Add map to Show page when geospatial metadata is available


ggeisler:
Below are some very rough ideas for this feature. I think we'll need to iterate on this design a bit because I have a few questions and it'll be good to get input and ideas from others, especially @mejackreed and @drh-stanford on possibilities and practicalities for the map-specific aspects. 

If I understand correctly, there are two main types of geospatial metadata we can expect to find in item records: bounding box (metadata is either degrees and minutes or box coordinates) or point (metadata is either lat/long or a geo name). In either case, the mockups below assume that, given an item with geospatial metadata, we generate a map for the appropriate type of data and put it at the top of the right column (at desktop viewport size) on the item show page, above the item metadata.

Details about the map area:
* In the mockup I tried to balance having the map large enough to read, but not dominate the right column. This is trickier than I expected. It seems like making the map small enough to fit here can also result in text on the map being difficult to read. I used a 2:1 aspect ratio for the map size, to use most of all of the column width while minimizing height.

* It feels like we need some sort of caption below the map, otherwise users might not know what the map is doing there. I'm not sure that ```Location of item's geospatial metadata``` is the best wording, though?

* Because the map is fairly small, users might want to view it larger to really understand the context of the location. I added links for that purpose (I'm not sure we want to link the map itself, because that could make it too easy to inadvertently click it and be take away from the page). Does this make sense, and if so, what are the source(s) most reasonable to link to?

### Desktop layout

I think we can use the same styling for viewports 992px and up. The version shown below is 1200px but the map still fits fine from 992 to 1199px.

![layout-desktop](https://cloud.githubusercontent.com/assets/101482/22170106/4eb78138-df37-11e6-990e-d8974704c87c.png)

### Mobile layout

Below 992px, I just did some very minor CSS adjustments to have the map take the full width, to match what we're doing with the embed and metadata. The map is much bigger here, but I think that's okay? It fits with the existing content where each section takes full width of the viewport, and I'm not sure we gain anything by reducing its size (I don't see how we would use the extra horizontal space), other than making it feel less dominant.

![layout-mobile](https://cloud.githubusercontent.com/assets/101482/22170281/26eb09c0-df39-11e6-8490-8c50b7614629.png)

### Point version

The mockups above use the bounding box version of the map. Here's how a point version looks:

![point-version](https://cloud.githubusercontent.com/assets/101482/22170143/badd14e0-df37-11e6-9492-1213a826544a.png)

### Configuration

[...]

PR #429, opened 2017-02-02
Add leaflet map to the show page

cbeer:
screen shot 2017-02-07 at 4 57 50 pm

screen shot 2017-02-02 at 8 52 33 am

Fixes https://github.com/sul-dlss/exhibits/issues/421


cbeer:
That's fine with me. We can let github do it.

PR #430, opened 2017-02-02
Route and mount for okcomputer checks at /status

tallenaz:


config/routes.rb
```diff
@@ -1,7 +1,8 @@
 Exhibits::Application.routes.draw do
   # API compatible with is_it_working checks
   match "/is_it_working" => "ok_computer/ok_computer#index", via: [:get, :options]
-  mount OkComputer::Engine, at: "/is_it_working"
+  match "/status" => "ok_computer/ok_computer#index", via: [:get, :options]
```
cbeer: Do we need to explicitly do this here, or are we back on the happy path:

https://github.com/sportngin/okcomputer#changing-the-okcomputer-route

Issue #448, opened 2017-03-13
Implement Bibliography Service


jkeck:
See #441 for design

- [x] Add basic form, controller, model, ~and background job~
- [x] Add a Sync From Zotero service
- [x] Add background job to trigger sync
- [x] Inject the `Bibliography` section into the record view
- [x] Add `Services` link to admin menu
- [ ] Hook up sync

PR #452, opened 2017-03-14
[WIP] initial version of Zotero bibliography service

drh-stanford:
This PR is connected to #448. It implements a `BibliographyService` API that only has a Zotero backend. Take a look at the specs for usage -- the main entry point is `bibliography` and `inverted_index` (to iterate through the druids that have bibliographies).

NOTE: Marked WIP as:
1. I dumped all of the code into the `zotero_api.rb` file and we may want to reorganize this into `Bibliography` and `BibliographyItem`, etc.
1. It's not using fixture data yet -- it's pulling data from the Zotero API live from my little test library online

### DIFF
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -23,7 +23,7 @@ Metrics/BlockLength:
 # Offense count: 1
 # Configuration parameters: CountComments.
 Metrics/MethodLength:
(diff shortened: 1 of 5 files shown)


mejackreed:
Looks good as a starting point, merging into feature branch for further work.

PR #454, opened 2017-03-19
Add service to hookup Zotero API and API Configuration classes for sync

jkeck:


app/models/concerns/exhibit_extension.rb
```diff
@@ -7,4 +7,16 @@ module ExhibitExtension
   included do
     has_one :bibliography_service
   end
+
+  def sync_bibliography
+    SyncBibliographyServiceJob.perform_later(self)
+  end
+
+  def build_bibliography_service
```
jkeck: Just out of curiousity, were you able to determine if the `build_bibliography_service` was already generated for us by `ActiveRecord`? http://guides.rubyonrails.org/association_basics.html#methods-added-by-has-one

spec/models/zotero_api/client_spec.rb
```diff
@@ -0,0 +1,51 @@
+require 'rails_helper'
+
+describe ZoteroApi::Client do
+  include ResponseFixtures
+  # TODO: replace with fixtures, currently drh's public library
```
mejackreed: Yes it is

jkeck: So you're saying if I replace that ID w/ `abc` then the tests will fail?

mejackreed: Sorry no, you are right here. 🐶

.rubocop_todo.yml
```diff
@@ -1,57 +1,74 @@
...
 RSpec/ExampleLength:
   Exclude:
     - 'spec/features/bibliography_service_feature_spec.rb'
     - 'spec/models/spotlight/dor/indexer_spec.rb'
 
-# Offense count: 28
+# Offense count: 2
+# Configuration parameters: EnforcedStyle, SupportedStyles.
+# SupportedStyles: have_received, receive
+RSpec/MessageSpies:
```
cbeer: Is it possible to address some of these newly introduced violations?

app/assets/stylesheets/modules/bibliography.scss
```diff
@@ -0,0 +1,5 @@
+.show-document {
+  h3 {
+    color: $color-cardinal-red;
```
cbeer: Is this really `$color-cardinal-red`, or might it be `$brand-primary`?

app/controllers/catalog_controller.rb
```diff
@@ -61,6 +61,8 @@ class CatalogController < ApplicationController
     config.show.partials.insert(1, :osd_or_embed)
     config.show.tile_source_field = :content_metadata_image_iiif_info_ssm
 
+    config.show.partials.insert(3, :bibliography)
```
cbeer: (I don't remember what the partials list is, but...) is this being inserted in the middle of the list, or just appended to the end (perhaps suggesting we could do something like `config.show.partials << :bibliography`?

app/controllers/services_controller.rb
```diff
@@ -14,6 +14,7 @@ def edit
 
   def update
     if @bibliography_service.update(update_params)
+      @exhibit.sync_bibliography if @bibliography_service.api_settings_changed?
```
cbeer: Could this logic be pushed down to the model (as an `before/after_update`?)

app/models/zotero_api/bibliography.rb
```diff
@@ -0,0 +1,21 @@
...
+  # A complete Zotero bibliography
+  class Bibliography < Array
+    # @return [Bibliography] all items sorted by Author-Date
+    def sort_by_author_date
+      self.class.new(sort_by(&:to_author_date))
+    end
+
+    # @param [String] `key` the bibliography item's `key` value
+    # @return [BibliographyItem] the matching item, or `nil`
+    def find(key)
```
cbeer: Doesn't this mask `Enumerable#find`? (and, the `each` below it looks suspiciously like it should use that..)

app/models/zotero_api/bibliography.rb
```diff
@@ -0,0 +1,21 @@
...
+      each { |item| return item if item['key'] == key }
+      nil
+    end
+
+    # @return [String] HTML for the entire bibliography, sorted by Author-Date
+    def render
+      "<ul>#{sort_by_author_date.collect(&:to_html).join}</ul>"
```
cbeer: Producing HTML through concatenation in a model isn't awesome. I think there's some rails magic (using stuff in `ActiveModel::Model`, maybe?) we could use to push this into a partial. If that sounds appealing, I can try to remember the method name (we use it in `Blacklight::Document`, I know...)

app/models/zotero_api/bibliography_item.rb
```diff
@@ -0,0 +1,52 @@
...
+    #
+    def to_html
+      "<li>#{fetch('bib')}</li>"
+    end
+
+    # @return [Array<String>] the list of druids associated with this bibliography item
+    def druids
+      druids = []
+      tags.each do |t|
+        druid = t.match(/([a-z]{2}[0-9]{3}[a-z]{2}[0-9]{4})/) # parse out druid from tag
```
cbeer: Can we pull this druid regex up to some configuration? There's some discussion about whether this pattern will change with SDR3 and it'll be easier to change if it isn't coded here.

app/models/zotero_api/bibliography.rb
mejackreed: @cbeer agree here that this isn't optimal. This is getting indexed into a solr document (and not displayed directly through the view).

Would you suggestion calling render partial methods in an indexing service then here?

app/models/zotero_api/bibliography.rb
mejackreed: Yeah this was due to pulling in previous changes. I'm not sure we even need it, going to nix it.

mejackreed: @jkeck ?? what do you think?

app/models/zotero_api/bibliography_item.rb
mejackreed: 06c89b9 feels a bit suboptimal, since we can't really put a regex in yaml

app/models/zotero_api/bibliography.rb
mejackreed: see 	cf9d0a9

app/controllers/services_controller.rb
jkeck: It seems like something with such dramatic side-effects like this in an ActiveRecord callback could be problematic.

This would make it impossible to update the model without syncing the service though, right? (without explicitly preventing callbacks, that is)

mejackreed: Yeah I came to a similar conclusion and felt like it would need to jump through some hoops. Maybe another model? But ... /shrug

app/models/zotero_api/bibliography.rb
cbeer: Nah, there's some way you get to say something like "when you call `render @bibliography`, go render this partial". I'll dig it up.

app/controllers/services_controller.rb
cbeer: ok. what about something stupid like `update_and_sync_bibliography`?

app/models/zotero_api/bibliography.rb
cbeer: https://robots.thoughtbot.com/rendering-collections-in-rails

app/controllers/services_controller.rb
jkeck: I would be 👌 with `update_and_sync_bibliography`.  I'll make the change unless @mejackreed beats me to it.

app/controllers/services_controller.rb
```diff
@@ -22,20 +23,22 @@ def update
...
+
   private
 
   def build_resource
-    @bibliography_service = if @exhibit.bibliography_service
-                              @exhibit.bibliography_service
-                            else
-                              ::BibliographyService.new(exhibit_id: @exhibit.id)
-                            end
+    @bibliography_service = BibliographyService.find_or_initialize_by(exhibit_id: @exhibit.id)
```
jkeck: What would you think about changing this to:

```
@bibliography_service = @exhibit.bibliography_service || @exhibit.build_bibliography_service
```

That way, the controller doesn't have to know that a `BibliographyService` class exists.

app/models/zotero_api/bibliography.rb
mejackreed: Ok. cool. I think I'm on the right path, and this seems much better with Rails 5 
https://medium.com/evil-martians/new-feature-in-rails-5-render-views-outside-of-actions-2fc1181e86a8#.omceel1px

mejackreed:
Had some issues getting this to work on stage this AM.. will try and troubleshoot when I can. cc @jkeck 

```log
2017-04-03T13:42:28.409Z 59330 TID-ngjio WARN: NoMethodError: undefined method `[]' for nil:NilClass
2017-04-03T13:42:28.409Z 59330 TID-ngjio WARN: /opt/app/exhibits/exhibits/releases/20170403122719/app/models/zotero_api/bibliography_item.rb:13:in `author'
/opt/app/exhibits/exhibits/releases/20170403122719/app/models/zotero_api/bibliography_item.rb:7:in `to_author_date'
/opt/app/exhibits/exhibits/releases/20170403122719/app/models/zotero_api/bibliography.rb:6:in `each'
/opt/app/exhibits/exhibits/releases/20170403122719/app/models/zotero_api/bibliography.rb:6:in `sort_by'
/opt/app/exhibits/exhibits/releases/20170403122719/app/models/zotero_api/bibliography.rb:6:in `sort_by_author_date'
```

mejackreed:
New issues that have come up when running on stage:

 - [ ] data column on sidecar is too small. https://app.honeybadger.io/projects/49092/faults/33193090
 - [ ] connection pooling performance for mysql https://app.honeybadger.io/projects/49092/faults/32421624
 - [ ] look into issue where field for solr is too big https://app.honeybadger.io/projects/49092/faults/33193484

app/models/zotero_api/bibliography_item_creator.rb
```diff
@@ -0,0 +1,29 @@
...
+
+    def first_name
+      metadata.fetch('firstName', nil)
+    end
+
+    def last_name
+      metadata.fetch('lastName', nil)
+    end
+
+    def formatted_author
```
drh-stanford: maybe use `to_s` instead?

config/settings.yml
```diff
@@ -8,3 +8,6 @@ stacks:
 
 purl:
   iiif_manifest_url: 'https://purl.stanford.edu/%{druid}/iiif/manifest.json'
+zotero_api:
+  endpoint: 'https://api.zotero.org/%{type}s/%{id}'
+  solr_document_field: zotero_bibliography_ssi
```
drh-stanford: may need to be a text field as per Honeybadger exception...

mejackreed: Yep, I'm not sure if there are any spotlight considerations with it being a text field? cc @jkeck ?

jkeck: Not that I'm aware of.  Text should be 👌

drh-stanford:
@mejackreed should we wait to merge this until the mysqld performance problems are fixed and the data column size is increased? or shall we go ahead with this merge and create tickets for those issues?

mejackreed:
@drh-stanford I think we are really close here. The migration was run successfully already on stage and the last sync job successfully completed but the sul-solr schema modifications have not seemed to percolate through. Is there something additional we need to do to get that going?

I think that is the final piece needed in place to 🚢

drh-stanford:
Yeah I thought that merging into `master` for sul solr configs sent those changes automagically to the solr cloud. But it sounds like we need to manually deploy those changes to the solr cloud -- I don't know how to do that.

PR #458, opened 2017-03-30
Remove medium index field from configuration…

### LINKED ISSUE #455: Index field 'medium' does not exist
https://github.com/sul-dlss/exhibits/blob/master/app/controllers/catalog_controller.rb#L146 should probably be `medium_display`?

jkeck:
…(we do have that field in our index).
Closes #455 

Looking over https://github.com/sul-dlss/sul-solr-configs/tree/master/exhibits it doesn't appear that we populate a `medium_display` (and we don't have `*_display` dynamic fields) and there is no other `medium_*` dynamic field referenced anywhere that I could find.

If we populate it in the indexing code, please feel free to chime in and I'm happy to update this PR to use that field instead.

### DIFF
--- a/app/controllers/catalog_controller.rb
+++ b/app/controllers/catalog_controller.rb
@@ -143,7 +143,6 @@ class CatalogController < ApplicationController
     config.add_index_field 'collector_ssim', label: 'Collector'
     config.add_index_field 'author_corp_display', label: 'Corporate Author'
     config.add_index_field 'author_meeting_display', label: 'Meeting Author'
-    config.add_index_field 'medium', label: 'Medium'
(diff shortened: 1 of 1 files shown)


cbeer:
Oh, very interesting. Does the same hold true upstream in https://github.com/sul-dlss/SearchWorks/blob/master/app/controllers/catalog_controller.rb#L161?

jkeck:
I believe that is true. The SearchWorks show views make little use of fields from the index these days.  I took a quick look over the SearchWorks schema and I'm not seeing any explicit `medium_display` or  dynamic `*_display` fields there either.

cbeer:
👍

PR #459, opened 2017-03-30
Add sort fields for uploaded resources

### LINKED ISSUE #457: Uploaded items should populate title sort fields
Uploaded items currently populate display and search fields:

https://github.com/sul-dlss/exhibits/blob/master/config/initializers/spotlight_initializer.rb#L1

It should also populate our `title_sort` (and other?) fields.

jkeck:
Closes #457 

I kept this in two commits in case we didn't want to override the entire `upload_fields` config just to add `date_sort`.

I can either yank that commit or squash these if we're 🆗 with it.


mejackreed:
👍  Looks good to me, but I'm unaware of any potential consequences. @cbeer do you have an opinion?

PR #468, opened 2017-04-10
use two unix processes for sidekiq with 5 threads each in stage

drh-stanford:
This PR makes 20 total threads and a maximum of 10 jobs for stage only

### DIFF
--- a/config/deploy/stage.rb
+++ b/config/deploy/stage.rb
@@ -6,4 +6,4 @@ Capistrano::OneTimeKey.generate_one_time_key!
 set :rails_env, 'production'
 
 set :sidekiq_role, :background
-set :sidekiq_processes, 10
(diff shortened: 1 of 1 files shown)


config/deploy/prod.rb
```diff
@@ -6,4 +6,4 @@
 set :rails_env, 'production'
 
 set :sidekiq_role, :background
-set :sidekiq_processes, 10
+set :sidekiq_processes, 2
```
mejackreed: I'm concerned that this change has been tuned somehow for production. Initially implemented here? https://github.com/sul-dlss/exhibits/commit/0d706d10855bb9f52a0f06dfbd26eb5a2a9bf214

Can we make the change in stage without doing it in prod yet?

drh-stanford:
nixed the prod changes

Issue #481, opened 2017-05-19
Determine language/copy for viewer config page


csbailey5t:
See this branch: https://github.com/sul-dlss/exhibits/tree/add_mirador_as_viewer

The feature (but not tests) to add Mirador as a viewer instead of the standard SUL-Embed is added. Following the template of the bibliography service, we add a menu option for the exhibit to get to a page to choose between Mirador and SUL-Embed with a radio button. We need to clarify the text for the exhibit administrator. 

Here's the current state: 
<img width="1114" alt="screenshot 2017-05-19 09 06 18" src="https://cloud.githubusercontent.com/assets/4672869/26256625/77dd28c2-3c72-11e7-99e9-d092db1b7109.png">

ggeisler:
@csbailey5t: This looks good. I have some minor suggestions below. Feel free to override anything I suggest that doesn't make sense to you.

![configuration_-_viewers___mirador_feature_text_-_stanford_university_libraries](https://cloud.githubusercontent.com/assets/101482/26271618/2a29d03a-3cbd-11e7-83b7-e84f2e4aff6d.png)

Here's what I've updated:

* I changed the section title to "Item Detail Page" because having "Viewers" as a heading twice is not ideal and if I understand correctly, this configuration is specific to the Item Detail page (and not, for example, the Embed widget), so it doesn't apply to all image resources displayed in the app. Plus, if we ever have more viewer configurations we can use tabs to separate them and turn that heading into the tab heading.

* Minor rewording of the instructions text:

> By default, the Item Detail page displays image resources using SUL-Embed as the viewer. You can change this setting to use a pre-configured version of Mirador as the viewer instead. The Mirador viewer provides minimal options (similar to 'zen mode'), but does include annotation viewing for any IIIF manifests with annotations.

* I capitalized the "embed" part of "SUL-Embed" for the radio button choice, just to be consistent with how we capitalized it in the instructions text.

Also, your screenshot above doesn't reflect this, but when I ran the branch locally neither radio button was selected. We definitely want to show the SUL-Embed button as the default selected choice, if it isn't already.

csbailey5t:
@ggeisler All of these make sense. Thanks for the review!

PR #483, opened 2017-06-07
Add mirador as viewer

### LINKED ISSUE #414: Mirador as viewer on show page
For certain exhibits, we want the option to change the default viewer on the record page to Mirador.  It should be configured in zen mode, with minimal chrome, no ability to change the image or add comparison panels.  It should have the ability to display, but not make annotations.

@blalbrit please clarify.

csbailey5t:
Fixes #414 and #481


app/views/catalog/_osd_or_embed_default.html.erb
```diff
@@ -1,5 +1,8 @@
+<% @viewer = Viewer.find_by(exhibit: @exhibit) %>
```
jkeck: Can we make this a local variable?  If it's needed by any other partials, we should pass it as a local.

app/views/catalog/_osd_or_embed_default.html.erb
```diff
@@ -1,5 +1,8 @@
...
 <% if document.uploaded_resource? %>
   <%= render partial: "openseadragon_default", locals: {document: document} %>
+<% elsif @viewer.viewer_type == "mirador" %>
+  <%= render partial: "viewers/mirador", locals: {document: document} %>
 <% else %>
   <%= render partial: "oembed_default", locals: {document: document} %>
```
jkeck: Another possible solution to this could be to implement `to_partial_path` on the `Viewer` class.

That method could be something along the lines of:

```ruby
def to_partial_path
  return 'oembed_default' if viewer_type == 'SUL Embed'
  "viewers/#{viewer_type}"
end
```

Then the view can just have

```
<% else %>
  <%= render @viewer, document: document %>
<% end %>
```

That way, we can add additional viewers w/o needing to update this partial each time.

app/views/catalog/_osd_or_embed_default.html.erb
```diff
@@ -1,5 +1,8 @@
...
 <% end %>
```
jkeck: Should this file name be changed now?

app/views/catalog/_osd_or_embed_default.html.erb
csbailey5t: @jkeck Since the partial seems to be set through config in the catalog_controller (https://github.com/sul-dlss/exhibits/blob/add_mirador_as_viewer/app/controllers/catalog_controller.rb#L61-L73), I'm not sure how one would pass a local to this partial. Any advice?

jkeck: @csbailey5t Is the `@viewer` variable set here used by another partial somewhere else?  If you change this from `@viewer` to `viewer` (and change the line below) do things break elsewhere?

app/models/viewer.rb
```diff
@@ -0,0 +1,10 @@
+##
+# Model for viewer
+class Viewer < ActiveRecord::Base
+  belongs_to :exhibit, class_name: Spotlight::Exhibit
+
+  def to_partial_path
+    return 'oembed_default' if viewer_type == 'sul-embed'
+    "../viewers/#{viewer_type}"
```
jkeck: I'm surprised that we need the `../` here.  Did `"viewers/#{viewer_type}" or `"viewers/_#{viewer_type}"` not work?

csbailey5t: Nope. It looked for the partials at `spotlight/...`, which is at the same level as `viewers`

app/views/catalog/_osd_or_embed_default.html.erb
csbailey5t: Seems to all work fine as `viewer` instead of `@viewer`.

jkeck: Can we use `viewer` instead then?  `@viewer` would indicate to me that we're setting a variable that is intended to have a greater scope than just this partial, where `viewer` would indicate that the scope is just to this partial (or wherever passed explicitly as a parameter or local).

csbailey5t: Sure. I'll make the change and push it up.

PR #487, opened 2017-06-19
Remove old demo environment

cbeer:


jkeck:
We're seeing this in other PRs too (#483).  Not sure exactly what's going on.

It seems odd that the build is only running 8 tests though (there are certainly more than that in this codebase).  Will try to do some digging (unless somebody else beats me to it).

PR #501, opened 2017-07-11
adds a feature flag to enable/disable sync bibliography service. disa…

mejackreed:
…bled on production

### DIFF
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -13,3 +13,5 @@ zotero_api:
   solr_document_field: zotero_bibliography_tsi
 sidekiq:
   logger_level: 'Logger::INFO' # The default from Sidekiq
+sync_bibliography_service:
+  enabled: false # FeatureFlag: SyncBibliograhyService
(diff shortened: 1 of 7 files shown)


config/settings.yml
```diff
@@ -13,3 +13,5 @@ zotero_api:
   solr_document_field: zotero_bibliography_tsi
 sidekiq:
   logger_level: 'Logger::INFO' # The default from Sidekiq
+sync_bibliography_service:
+  enabled: true # FeatureFlag: SyncBibliograhyService
```
jkeck: Any thoughts on setting this to `false` by default and setting it to `true` in test/dev?

The only reason I mention that, is that this (confusingly) checked in file will get overrwritten by what's in [`shared_configs`](https://github.com/sul-dlss/shared_configs/blob/exhibits-prod/config/settings/production.yml) and could accidentally lead us to releasing the feature in production.

mejackreed: Great suggestion, I've updated accordingly.

PR #509, opened 2017-07-14
[DONT MERGE see 515] Revert "remove sidekiq-statistics"

mejackreed:
Reverts sul-dlss/exhibits#506

This is causing some issues deployed. Will revert to look into later.


mejackreed:
See https://github.com/sul-dlss/exhibits/pull/516 instead

PR #515, opened 2017-08-02
remove sidekiq_web status

mejackreed:


mejackreed:
See https://github.com/sul-dlss/exhibits/pull/516 instead

PR #594, opened 2017-09-25
Additional development / testing README updates

camillevilla:
This PR:
- consolidates development instructions 
- adds the appropriate task for running tests

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,7 +30,7 @@ Exhibits need to provide the following configuration files:
 
 ## Reindexing content
 
-A rake task is provided to (re)index content into the Solr index. It uses the configured sets in `config/exhibit.yml`.
+A Rake task is provided to (re)index content into the Solr index. It uses the configured sets in `config/exhibit.yml`.
 
 ```console
 $ rake spotlight:index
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -46,26 +46,19 @@ $ bundle install
...
-```
-
-Seed test fixtures and create an admin user (e.g. archivist1@eample.com):
-```console
-$ bundle exec rake spotlight:seed
-$ bundle exec rake spotlight:initialize
-```
-
-Spin up the rails server. (`REMOTE_USER` should match the name of the user you create when prompted. This will allow you to bypass the webauth authentication.)
+You can spin up the rails server, jetty, and populate the solr index using this command:
```
mejackreed: `rails s` is just running the rails server I believe, right?

Also if we keep a comment about a command that does those several things, could we change jetty -> solr_wrapper ? (jetty is a historical reference) A similar change here: https://github.com/sul-dlss/exhibits/pull/581/files#diff-04c6e90faac2675aa89e2176d2eec7d8R43 (which I'm willing to close)

camillevilla: ✅

camillevilla:
@mejackreed good catch. I've made requested changes.

PR #598, opened 2017-09-27
update Mirador config to show sidepanel and enable layers, annotation…

aeschylus:
…s, and TOC


aeschylus:
mirador sidepanel

aeschylus:
If we want the nicer styling from the recent release, we would need to update to the mirador_rails gem.

Issue #600, opened 2017-09-27
add ability to retrieve BibTeX data with each bibliography item


drh-stanford:
To simplify the bibliography rendering (see #585), it's useful to have BibTeX data for the `citeproc` gem. The `bibtex-ruby` gem provides a mapping between BibTeX and the citeproc schema (`to_citeproc`). Since we're exporting the bibliography as BibTeX, that's the most straightforward way to implement rendering.

PR #601, opened 2017-09-27
Indexes bibliography documents

### LINKED ISSUE #583: Ability to index a bibliographic item from a fixture
Split out from #572 

Just index something quick and basic.

### LINKED ISSUE #595: Add spotlight resource indexer for bibliography items
Split out of work in #583 originally.

Add in a custom Spotlight Resource indexer see: https://github.com/projectblacklight/spotlight/wiki/Resource-Scenarios#on-demand-resource-indexing

mejackreed:
This PR enables external resources (currently filenames of fixtures) to be indexed using Traject + Spotlight External Resources

parkerbib2

Things to finish before WIP removed:

 - [x] Feature flag it (based on what @jkeck sets up) at least the initializer add to form
 - [x] Test it (not sure how to do this).. ideas?

Closes #595, Closes #583 

Known issues/ needed enhancements (likely to be split out issues):

 -  Have the exhibit specific fields be added (not sure why this isn't happening but the relevant magic in Spotlight is here: https://github.com/projectblacklight/spotlight/blob/master/app/services/spotlight/solr_document_builder.rb#L13-L23 Split out ticket #607
 -  More metadata! - this will be driven by design requirements
 - Store bibtex? cc @drh-stanford #600
 - Enhance how the `.bib` file gets in place see #611


jkeck:
Taking a look through now.  One comment from the gif (and from the `Known issues/ needed enhancement`) is the expectation that there is a matching filename uploaded to the server already (in this case `article.bib`)?

spec/features/bibliography_resource_integration_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+RSpec.describe 'Bibliography resource integration test', type: :feature do
+  subject(:bibliograpy_resource) { BibliographyResource.new(url: 'article.bib', exhibit: exhibit) }
+
+  let(:exhibit) { FactoryGirl.create(:exhibit) }
+  let(:title_fields) do
+    %w(title_245a_search title_245_search title_sort title_display title_full_display)
+  end
+
+  it 'can write the document to solr' do
+    bibliograpy_resource.reindex
```
drh-stanford: this is a bit unclear why there's not an `expect` assertion here

mejackreed:
@jkeck most definitely. I've create #611 to address that.

spec/features/bibliography_resource_integration_spec.rb
mejackreed: Gotcha I guess it is just so that an error doesn't raise.

drh-stanford: in that case, i think you can do `expect { bibliography_resource.reindex }.not_to raise_error`

mejackreed: Of course, good catch. Added.

mejackreed:
@drh-stanford Indexing with those explicit spotlight fields is accomplished in #606. Specifically: fc82b1defb5729376ce9fa66cb7e0f2f9e14c35d

config/routes.rb
```diff
@@ -31,6 +31,7 @@
 
   resources :exhibits, path: '/', only: [] do
     resource :dor_harvester, controller: :"dor_harvester", only: [:create, :update]
+    resource :bibliography_resources, only: [:create, :update]
```
jkeck: Probably less important, but wondering if we should put the route behind the feature flag as well.

mejackreed: We could.. but I didn't think it was really necessary. Happy to make that change if you think we should.

app/controllers/bibliography_resources_controller.rb
```diff
@@ -0,0 +1,31 @@
...
+  before_action :build_resource
+  authorize_resource
+
+  def create
+    @resource.update(resource_params)
+
+    if @resource.save_and_index
+      redirect_to spotlight.admin_exhibit_catalog_path(current_exhibit)
+    else
+      redirect_to spotlight.new_exhibit_resource_path(current_exhibit)
```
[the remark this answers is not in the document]
mejackreed: Great suggestion. Added a ticket for it here: #614

PR #610, opened 2017-09-28
closes #522, Mirador height fix

### LINKED ISSUE #522: Re-size embedded Mirador viewer to match what it was previously
Current work-around for Mirador cut the size to something that is largely unusable for the Parker case. Please re-size to the same size/page-percentage that it was previously.

aeschylus:


aeschylus:
This fixes it, but using vh on an iframe somehow causes the units to be in percent of the iframe itself instead of the overall viewport. There is one more way I would like to try fixing this in a way that doesn't require this "900vh" hack, which probably won't work well on all browsers: I think wrapping the iframe in yet another container should allow that container to take the vh unit magnitude from the correct viewport element (vh is calculated from the _first containing element_ within the viewport).

app/assets/stylesheets/base.scss
```diff
@@ -1,4 +1,3 @@
-@import "bootstrap-tagsinput";
```
mejackreed: Is removal of this a related change?

aeschylus:
Okay. This is ready to merge whenever others would like.

app/assets/stylesheets/base.scss
aeschylus: Hm. This isn't related. It may have been accidentally removed while resolving conflicts when I rebased on master. Let me add this back in.

aeschylus: @mejackreed Okay. This has been fixed.

Issue #603, opened 2017-09-27
Remove first pass at BibliographyService


mejackreed:
Remove the work added to Exhibits that culminated in https://github.com/sul-dlss/exhibits/pull/477

A revert might be hard.

PR #619, opened 2017-09-28
skip legacy bib service specs

mejackreed:
Trying to fix a travis glitch


drh-stanford:
These tests will be removed later under #603

PR #621, opened 2017-09-28
add BibTeX and formatted bibliography to ingest pipeline

### LINKED ISSUE #600: add ability to retrieve BibTeX data with each bibliography item

drh-stanford:
This PR fixes #600. Use `document.bibtex` to get a `BibTeX::Bibliography`. Also adds `document.reference?` and `document.formatted_bibliography`.

Note that you will need to do a `solr_wrapper clean` for this PR to work as the `schema.xml` file has a change. The PR uses a Solr `text` field without indexing to store the BibTeX and HTML data.

### DIFF
--- a/app/models/concerns/bibliography_concern.rb
+++ b/app/models/concerns/bibliography_concern.rb
@@ -5,4 +5,17 @@ module BibliographyConcern
   def bibliography
     fetch(Settings.zotero_api.solr_document_field, nil)
   end
+
+  def reference?
+    fetch('format_main_ssim', []).first == 'Reference'
+  end
+
+  def bibtex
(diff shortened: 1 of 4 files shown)


drh-stanford:
The coveralls went down because we're no longer testing the old bibliography stuff. Not sure why it affects this PR...

Issue #611, opened 2017-09-28
Enhance how the .bib file gets in place


mejackreed:
Split out of #601 

We need to make this more adoptable for a user (besides just opening a file in `spec/fixtures/bibliography`

mejackreed:
Could be a file upload or a url reference.

PR #624, opened 2017-09-28
Add flash messages for bibliography resource creation

### LINKED ISSUE #614: Add flash messages for bibliography resource creation
Per comment here: https://github.com/sul-dlss/exhibits/pull/601#discussion_r141672929

camillevilla:
Closes #614 

- This PR adds a flash message for successful creation of bibliography resources via the `Bibliography via BibTeX`. 
- Adds an error message for unsuccessful creation (`There was a problem with adding the BibTeX resource.`)
But...we don't have any validations on the form yet. @mejackreed should this be added as a separate ticket or will this be part of #611?

## Screenshot
successful_flash_message

### DIFF
--- a/app/controllers/bibliography_resources_controller.rb
+++ b/app/controllers/bibliography_resources_controller.rb
@@ -12,9 +12,11 @@ class BibliographyResourcesController < Spotlight::ResourcesController
     @resource.update(resource_params)
 
     if @resource.save_and_index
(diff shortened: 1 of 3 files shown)


mejackreed:
I think the validations should be part of #611 like you suggest. Since we don't know what we are doing yet, lets leave it open.

Issue #585, opened 2017-09-25
ability to provide citations of these first-class objects in a manuscript item view display, aggregating all bibliographic objects "about" a specific manuscript item


mejackreed:
Splitout from #572. Depends on #589, and #600.

See https://github.com/sul-dlss/exhibits/issues/578#issuecomment-332332320 for background

- [x] Integrate citeproc gem to implement reference style (Chicago, Author-Date, ascending) (PR #602)
- [x] Handle embedded LaTeX in bibliography items (`{}`, `\textit{}`, etc.) (PR #602)
- [x] Query Solr for all bibliography items relating to given manuscript, needs to be gated by exhibit (requires #589)
- [x] Fetch BibTeX for all the relevant bibliography items (requires #600)
- [ ] Add styled, sorted bibliography to show page (#576)

mejackreed:
Dependent on #578 and #583

mejackreed:
Potentially will require an asynchronously lookup to Solr.

drh-stanford:
@blalbrit the Parker bibliography has LaTeX markup in the titles (and possibly elsewhere). For example, `This is \textit{my {Capitalized Title}}`. Is this a consistent practice for the Parker folks? When rendering the bibliography can they be ignored, or should we be rendering them as HTML?

blalbrit:
@drh-stanford Weird. The Parker folks aren't doing that (they were just entering text into Endnote and FileMakerPro) - must be an artifact from some export/import process. Those can be ignored, I think. Can you tell me what Zotero record that example is from? I'll check the source data just to make sure.

drh-stanford:
@blalbrit it looks like they are putting HTML markup in the records. The BibTeX export must be converting it into TeX commands.

See (Whatley 1986) -- there is HTML in the title.

blalbrit:
@drh-stanford Great - thanks for the example. That is an artifact of the EndNote export. Original looks like:
<img width="294" alt="screenshot 2017-09-27 14 51 24" src="https://user-images.githubusercontent.com/796990/30939772-7b3e62c8-a393-11e7-90bd-95a2ccc7eadf.png">
Please ignore that mark-up for the work that you all are doing.

drh-stanford:
Currently blocked by #589 and #600.

PR #626, opened 2017-09-28
add related_documents to bibliography resources

### LINKED ISSUE #589: Relate indexed bibliography object to indexed druid parent

drh-stanford:
This PR closes #589 by adding the data required to do manuscript -> references queries.

Marked WIP:
- [x] Add gated query (i.e., filtered by exhibit) to `related_documents` method (split into https://github.com/sul-dlss/exhibits/issues/632)
- [x] Add ability to retrieve `formatted_bibliography` from non-Reference resources (maybe do as part of #585) (split into https://github.com/sul-dlss/exhibits/issues/632)
- [x] Should we add validation that related druids **must** exist in the exhibit at index-time? (split into https://github.com/sul-dlss/exhibits/issues/634)
  - Note we can validate using `Spotlight::Exhibit.find(sidecar.exhibit_id).slug` matching the Traject indexer's `context.settings.exhibit_slug`

### DIFF
--- a/app/models/concerns/bibliography_concern.rb
+++ b/app/models/concerns/bibliography_concern.rb
@@ -16,6 +16,5 @@ module BibliographyConcern
 
   def formatted_bibliography
     fetch('formatted_bibliography_ts', []).first if reference?
-    # TODO: for non-reference resources we'll generate a bibliography for them
   end
 end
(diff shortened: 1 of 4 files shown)


drh-stanford:
The coveralls drop is because this PR disables the old Zotero API testing which was causing problems -- we probably should do #603 soon.

mejackreed:
I've split out a ticket for adding the test #640 . We need this to fix the build .

drh-stanford:
The testing for this is deferred until https://github.com/sul-dlss/exhibits/issues/632 when I can build a feature test for it.

Issue #605, opened 2017-09-27
Design for Show page for Bibliography Item


ggeisler:
I think we decided to make bib items first-class objects, and we do have a ticket for including bib items in search results (#575), but unless I've missed it I don't think we have a ticket for a bib item show page.

Below is a first-pass at a design. This is the show page the user would arrive at when clicking on a bib item in the search results, or from the Bibliography list on the show page for a document (manuscript).

(Note that the data shown in the mockup is not accurate in terms of the real documents associated with the bib entry shown.)

![spotlight_graffle__parker__public___bib_item_show_page](https://user-images.githubusercontent.com/101482/30941698-9c63fbcc-a39b-11e7-9105-c25d19436dd7.png)

mejackreed:
We may also have the ability to provide a download of the raw `BibTeX` data for users

ggeisler:
Below is an update, putting the notes into the metadata list. I also added the download button as @mejackreed suggested (I started mocking up a copy to clipboard option, which seems potentially useful, but much more involved to implement, so decided to leave it off, given we haven't heard a real requirement for it).

![bib-show-page-v2](https://user-images.githubusercontent.com/101482/30987009-58491de6-a44a-11e7-96b8-962c8e1971f5.png)

ggeisler:
Minor update, just adds the "View on Zotero site" button, which would link to the reference item's page on the Zotero site.

![bib-show-page-v3](https://user-images.githubusercontent.com/101482/31025861-16088ade-a4f9-11e7-8d72-5ae57d2a6c07.png)

aeschylus:
This depends on having a meeting to refine the bibTex fields.

See:

drh-stanford:
See #589 for how to get the related documents to the given reference

camillevilla:
Per dicussion with @blalbrit on 10/3, `Item Type` should be renamed `Reference Type` (e.g. Book, Journal Article). `Resource` type will be  `Reference`.

camillevilla:
This design ticket will be closed. Implementation tickets will refer back to this.

drh-stanford:
@ggeisler here's what the View on Zotero site button looks like when the Edit button is enabled:

![screen shot 2017-10-05 at 12 11 34 pm](https://user-images.githubusercontent.com/1861171/31247790-8748638c-a9c6-11e7-9044-8dad51cc0465.png)

[...]

PR #628, opened 2017-09-29
Add author and publication to show page

camillevilla:
Fixes part of #605 

This PR adds `author_person_full_display` and `pub_display` to the show page. I followed the pattern from #622. I wanted to ask some questions at stand-up before adding fields to `catalog_controller.rb`.

## Before
before_show

## After
after_show

### DIFF
--- a/lib/traject/bibtex_config.rb
+++ b/lib/traject/bibtex_config.rb
@@ -22,6 +22,14 @@ to_field 'title_uniform_search', lambda { |record, accumulator, _context|
   accumulator << record.title.to_s(filter: :latex)
 }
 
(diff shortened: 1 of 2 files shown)


lib/traject/bibtex_config.rb
```diff
@@ -22,6 +22,15 @@
   accumulator << record.title.to_s(filter: :latex)
 }
 
+to_field 'author_person_full_display', lambda { |record, accumulator, _context|
+  # byebug
```
mejackreed: Can we drop this comment?

camillevilla: ✅

Issue #632, opened 2017-09-29
Load the sorted bibliography asynchonously on manuscript show page


drh-stanford:
This PR is split out from #589 and #585. Depends on #626 

- [x] Implement a way to use `CiteProc` to render a sorted bibliography (`formatted_bibliography_ts`)
- [x] Implement a query API to retrieve all bibliography references for a given manuscript
- [ ] Render sorted bibliography asynchronously on the manuscript show page PR #660

A suggested implementation is to use JavaScript and the JSON API to get the list of formatted references (which will apply the exhibit gated query) for that manuscript, and apply an author-date-title sort based on Solr fields.

Here's the JSON API query needed to retrieve all of the references for a given resource (druid):

```
f[related_document_id_ssim][]=gs233db8425&
f[format_main_ssim][]=Reference&
sort=author_sort asc, pub_year_isi asc, title_sort asc&
rows=999&
format=json
```

The other advantage is that the injection of the links to Zotero is straight-forward -- see #539.

drh-stanford:
Blocked by design discussion needed for #635.

drh-stanford:
@ggeisler how does this look for the bibliography (ignore the bad sorting), based on the design in #576?

![screen shot 2017-10-03 at 12 44 08 pm](https://user-images.githubusercontent.com/1861171/31145493-b7952f96-a838-11e7-8256-2a042fc5d527.png)

drh-stanford:
I got the sorting to work and here's what the full manuscript show page looks like:

![screen shot 2017-10-03 at 2 27 19 pm](https://user-images.githubusercontent.com/1861171/31150046-14469636-a847-11e7-9b2e-12a7084c28df.png)

ggeisler:
This is pretty minor, but the `div` with the Bibliography section looks like it has some default left margin (~20px) that is preventing it from being left-aligned with the sections above it (e.g. the left-edge of the view and item title). 

Maybe try to target the whole bibliography section and add a `margin-left: 0;` to it? (Or don't worry about it and I'll deal with it later on.)

ggeisler:
Perfect @drh-stanford!

Issue #576, opened 2017-09-25
Design for manuscript show page including bibliographic items


blalbrit:
I believe @ggeisler has already done these. Link to those here?

blalbrit:
See #441

ggeisler:
The mockup below updates slightly the previous design to reflect recent comments in Slack from @blalbrit, specifically:
 
- The bibliography list entries are now formatted with a hanging indent
- The title of each bib entry is linked to a show page for that entry, as proposed in #605.

(Note that the data shown in the mockup is not accurate in terms of the real bibliography data for the document shown.)

![show-page-with-bib-items](https://user-images.githubusercontent.com/101482/30941775-055ca3b8-a39c-11e7-89a2-b740728871ab.png)

drh-stanford:
One concern I have is that these hyperlinks use the title as the anchor. It'll be some extra work to get that to render correctly I think. We may need to modify the CSL or somehow do a search and replace on the title. If there's another way to do the linking that makes sense, I'd be interested to hear.

ggeisler:
@drh-stanford Yeah, I wanted to present that as the preferred approach, but I realize that it might not be feasible to wrap the link around just the title portion of the reference.

If it isn't feasible, we can just add a distinct "View full reference" link after each bib item instead. It'll just make the list a bit more busy.

I can update the mockup to provide a more exact recommendation.

mejackreed:
We could also link them to a bibliography show page. And within a bib show page link all of the cited articles

ggeisler:
@mejackreed That's what they are linking to in the mockup, unless I'm misunderstanding your meaning. They link to the page mocked up here: https://github.com/sul-dlss/exhibits/issues/605

mejackreed:
Perfect @ggeisler sorry I misunderstood.

ggeisler:
@drh-stanford Here's an updated version of the mockup, with the links distinct from the reference itself. If the CSS needed to match the mockup gets tricky, I can work on that after the basic implementation is done.

![show-page-with-bib-items-v2](https://user-images.githubusercontent.com/101482/30984159-6ad137c2-a441-11e7-8c27-da5b7fd18f77.png)

PR #635, opened 2017-09-29
add controller for formating sorted bibliographies

drh-stanford:
This PR is connected to #632. It establishes a new route that takes in one or more bibliography document identifiers and returns a sorted bibliography in HTML.

For example, from http://localhost:3000/default/bibliography_formatting?id[]=QTWBAWKX&id[]=GWT7TMKR&id[]=KJ5WQIFZ&id[]=WJ72DTUD (note these IDs are from the full Parker bibliography)

screen shot 2017-10-02 at 11 14 06 am

whose HTML looks like: 

screen shot 2017-10-02 at 11 21 52 am

### DIFF
--- a/app/controllers/bibliography_formatting_controller.rb
+++ b/app/controllers/bibliography_formatting_controller.rb
@@ -1,0 +1,13 @@
+# frozen_string_literal: true
+
+##
+# Class for handling BibliographyResourcesController
+class BibliographyFormattingController < ApplicationController
+  def show
+    bibliography_ids = Array.wrap(params.require(:id))
+    bibtex = bibliography_ids.collect do |document_id|
+      SolrDocument.find(document_id).bibtex.to_s
+    end.join("\n")
+    render html: Bibliography.new(bibtex).to_html.html_safe # rubocop: disable Rails/OutputSafety
+  end
+end
--- a/app/models/concerns/bibliography_concern.rb
+++ b/app/models/concerns/bibliography_concern.rb
@@ -11,10 +11,10 @@ module BibliographyConcern
   end
 
   def bibtex
-    BibTeX.parse(fetch('bibtex_ts', []).first) if reference?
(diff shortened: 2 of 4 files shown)


mejackreed:
A general question: Does this approach enable us to add the design described here: https://github.com/sul-dlss/exhibits/issues/576#issuecomment-332926351

Specifically I'm wondering about the link insertion.

I also am interested about the public/non-public viewing with this approach and if gated discovery is possible.

drh-stanford:
Yes, for the async design in #632, there's 2 steps. (1) the JavaScript is responsible for querying via the exhibits JSON API. (2) It'll then pass all the bibliography IDs to this controller to retrieve the sorted, formatted bibliography.

I was thinking about the link insertion, and this PR doesn't handle it -- I was considering breaking out that requirement into a different ticket from #632 (and the uber ticket #585). We may need to have the controller be smarter about retaining the Zotero IDs in the formatted bibliography. That would likely involve modifying the CiteProc chicago-author-date style, or to wrap the title with a link before we pass it to CiteProc. The basic problem is that we defer to CiteProc to sort the bibliography, which is a non-trival operation afaik.

I wasn't aware that there's public vs non-public access to these resources. There's no admin/cancan stuff in the controller intentionally as I assume everything was public. The bibliography identifiers are not gated at this point in the controller. The controller accepts the IDs as they are.

mejackreed:
Gotcha. I was hoping that we could use some type of sort field and solr within the `CatalogController` to do the sorting. I can understand if that isn't possible, but I think we definitely need to apply the gated aspect to this feature.

drh-stanford:
Let's discuss this offline. The reason we're doing the JSON API in (1) is to do the gated discovery and there's confusion how to do this from within the Spotlight API.

drh-stanford:
After a design discussion we'll be taking a different approach that leverages Solr to do the author-date sorting algorithm

Issue #616, opened 2017-09-28
Index the complete Parker bibliography


drh-stanford:
There's 6k+ bibliography items in the Parker production library in Zotero. Right now only 3.5k are indexing correctly. We need to ensure that we can index them all. I put the BibTeX export in our shared folder here:

https://drive.google.com/open?id=0B5fLh2mc4FCbSTF3LWZRd3dpR0E

See #608 for background.

Handle cases:
 - [x] no title (skip)
 - [x] missing optional fields (empty)
 - [x] `Bibliography.new().to_html` raises error

mejackreed:
I'm getting an error which is failing the process:

```
2017-09-29T11:08:32.925Z 19910 TID-oxhjnwqho WARN: ArgumentError: wrong number of arguments (given 1, expected 0)
2017-09-29T11:08:32.926Z 19910 TID-oxhjnwqho WARN: lib/traject/bibtex_config.rb:15:in `to_s'
```

mejackreed:
Apparently 2 records do not have titles: http://zotero.org/groups/1051392/items/NNHKWP6C and http://zotero.org/groups/1051392/items/6Q6TF4HD 

This seems like bad data. @blalbrit ?

blalbrit:
@mejackreed not necessarily bad data, just currently incomplete. As they maintain the main bibliography, Parker staff occasionally come across partial references to publications which they then will track down later, verify, and fill out further.

blalbrit:
^^ two out of ~6K doesn't seem too bad. If they wind up being unverifiable, then these could be removed from the solr index (and Zotero) by me, correct? (that was the impression I got from discussion with @drh-stanford about the need for administrative intervention in very rare instances).

drh-stanford:
@blalbrit can we enforce that a bibliography needs to have a title to be indexed?

blalbrit:
I would be happy with that - even if the title is a placeholder, I can train the stakeholders on that @drh-stanford

drh-stanford:
An implementation note: You can use `context.skip!` in Traject to reject processing a given record. This may be better than validating all of the BibTeX records in the reader.

mejackreed:
@drh-stanford #637 handles some of this. Still getting an error on 

```
@article{http://zotero.org/groups/1051392/items/MT4A9V2I,
[...]

PR #637, opened 2017-09-29
Better support of BibTeX record -> Solr parsing to handle no titles a…

mejackreed:
…nd missing fields

Partial of #616 

~~Cherry-picked 7baa4fefea69406b631999091e3f5c07477646a9 so things can be 💚~~ shipped that PR now.


app/services/bibliography_builder.rb
```diff
@@ -39,6 +39,7 @@ def traject_indexer
   private
 
   def convert_id(doc)
+    return nil unless doc['id']
```
jkeck: `nil` is redundant in this return I believe.

mejackreed: Thanks for catching that. I've pushed another commit that updates it since #633 is relying on one of these commits now.

PR #639, opened 2017-10-02
add support for a bibtex file upload

### LINKED ISSUE #611: Enhance how the .bib file gets in place

mejackreed:
Closes #611 

bibupload

### DIFF
--- a/app/models/bibliography_resource.rb
+++ b/app/models/bibliography_resource.rb
@@ -2,4 +2,6 @@
 # BibliographyResource model class
 class BibliographyResource < Spotlight::Resource
   self.document_builder_class = BibliographyBuilder
+
+  store :data, accessors: [:bibtex_file]
 end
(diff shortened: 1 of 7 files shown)


app/models/bibliography_resource.rb
```diff
@@ -2,4 +2,6 @@
 # BibliographyResource model class
 class BibliographyResource < Spotlight::Resource
   self.document_builder_class = BibliographyBuilder
+
+  store :data, accessors: [:bibtex_file]
```
jkeck: Wondering if we would get some benefit using `CarrierWave` instead of storing the raw data here.

mejackreed: I've add an `accept: 'application/x-bibtex'` to the form, but I'd like to stay away from carrierwave at this time. It feels like maybe a level of indirection we don't need.

PR #642, opened 2017-10-02
remove legacy zotero bibliography sync

### LINKED ISSUE #603: Remove first pass at BibliographyService

mejackreed:
Closes #603


config/settings.yml
```diff
@@ -13,8 +13,6 @@ zotero_api:
   solr_document_field: zotero_bibliography_tsi
```
drh-stanford: these `zotero_api` settings can also be deleted

mejackreed: Good catch. Pushed

PR #646, opened 2017-10-02
improve BibTeX indexing when BibTeX.parse lexer throws warnings

drh-stanford:
This PR is connected to #616, and it fixes a bug/feature of the BibTeX indexer when it gets title markup that it cannot handle gracefully -- namely `{\textbackslash}` which is used in the Parker bibliography.


app/models/bibliography.rb
```diff
@@ -40,8 +40,8 @@ def render(format = 'html')
   # of cases.
   def remove_latex_markup!
     @bibliography.each do |item|
-      item.title = strip_latex(item.title)
-      item.booktitle = strip_latex(item.booktitle)
+      item.title = strip_latex(item.title) if item.respond_to?(:title)
+      item.booktitle = strip_latex(item.booktitle) if item.respond_to?(:booktitle)
```
jkeck: I'm wondering if this isn't a good opportunity for either a NullObject pattern or to not process these kinds of things in general?  I worry that we'll regularly be asking questions about the object's interface (which couples this code to tightly together).

drh-stanford: Good idea. I put that into a ticket -- https://github.com/sul-dlss/exhibits/issues/652

PR #653, opened 2017-10-02
add SolrDocument.related_document_ids and use SolrDocument.first where possible

### LINKED ISSUE #640: Add test for related documents for a bibliography

drh-stanford:
This PR closes #640 by adding tests of the Traject code that indexes related documents using the BibTeX.keywords field. Also, uses `first` method in the bibliography concern where possible. Finally, I cleaned up the indexing specs a bit.


drh-stanford:
Ok, the conflict is resolved.

PR #656, opened 2017-10-03
update fixtures to have all fields available from Zotero export

### LINKED ISSUE #649: Analysis: get a list of BibTeX fields supported by Zotero
Split off from #605
Meeting with @blalbrit @drh-stanford  and @camillevilla to discuss 10/3

drh-stanford:
This PR closes #649 and is connected to #605. It updates our bibliography fixture data to have all the fields that we want to render/search for Reference objects.


mejackreed:
How do we get those fields in there? Do we need to modify the bibtex export script?

drh-stanford:
We had to handcraft them as we need the superset of all fields. The fixtures aren't meant to be a round-trip for the export mechanism are they?

mejackreed:
Nope, but I was just curious about it. 👍

PR #660, opened 2017-10-03
load bibliography section asynchonously on manuscript show page

### LINKED ISSUE #539: For a bibliography item enable link to Zotero bibliographic item
As a user of Parker on the Web,
I want to be able to click on a bibliographic item and be linked to the full bibliographic item in Zotero online,
In order to be able to see additional notes and commentary provided by the library (which includes manuscript-specific notes).

Ok, it sounds like the work here is:

- [ ] Modify the bibliography show page to include the "View on Zotero site" button -- see PR #670.
- [x] Modify the bibliography on the manuscript show page to have a link to View in Zotero. This is done in #660.

### LINKED ISSUE #632: Load the sorted bibliography asynchonously on manuscript show page

drh-stanford:
This PR closes #632 and closes #539.

- [x] Async loading on manuscript (resource) show page
- [x] Proper sorting by Chicago-Author-Date
- [x] Automated testing
  - [x] View tests
  - [x] JavaScript tests

Note that to try this out, you need to (in this order):

  1. Load some Parker resources into your exhibit (e.g., cg531kv2466) 
  2. Load some BibTeX items into your exhibit that refer to the druids of your Parker objects (ones that reference cg531kv2466 in the keywords field)

### After

screen shot 2017-10-04 at 2 49 37 pm

### DIFF
--- a/app/assets/javascripts/bibliography.js
+++ b/app/assets/javascripts/bibliography.js
@@ -1,0 +1,41 @@
+/* eslint-disable camelcase */
+/* global Bibliography */
+
+(function (global) {
+  var Bibliography;
+
+  Bibliography = {
+    init: function (el) {
+      var $el = $(el);
+      var data = $el.data();
+      $.getJSON(data.path, {
+          'f[related_document_id_ssim][]': data.parentid,
+          'f[format_main_ssim][]': 'Reference',
+          sort: data.sort,
+          format: 'json',
+          rows: '1000'
+        }, function (response) {
+        for (var i = 0; i < response.response.docs.length; i++) {
+          var bibEntry = response.response.docs[i];
+          var html = '<p class="bibliography-body">' +
+                      bibEntry.formatted_bibliography_ts +
+                      ' <a href="' +
+                      data.baseurl + bibEntry.id +
+                      '">[View full reference]</a>' +
+                      '</p>';
+          $el.find('.bibliography-list').append(html);
+        }
+      });
+    }
+  };
+
+  global.Bibliography = Bibliography;
+}(this));
+
+Blacklight.onLoad(function () {
+  'use strict';
+
+  $('.bibliography-contents').each(function (i, element) {
+    Bibliography.init(element); // eslint-disable-line no-undef
+  });
+});
--- a/app/views/catalog/_bibliography_default.html.erb
+++ b/app/views/catalog/_bibliography_default.html.erb
@@ -1,0 +1,9 @@
+<div class='col-md-12 bibliography-contents'
+     data-path="<%= search_exhibit_catalog_path(exhibit_id: @exhibit) %>"
+     data-parentid="<%= @document.id %>"
+     data-baseurl="http://zotero.org/groups/1051392/items/"
+     data-sort="author_sort asc, pub_year_isi asc, title_sort asc">
+     <%# TODO: retrieve this sort configuration from the blacklight config %>
+     <h3 class="col-md-9">Bibliography</h3>
+     <div class="col-md-9 bibliography-list"></div>
+</div>
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -21,6 +21,7 @@
 //= require blacklight/blacklight
 
 //= require analytics
+//= require bibliography
 //= require blacklight_gallery
 //= require blacklight_heatmaps
 //= require blacklight_oembed
--- a/app/assets/stylesheets/modules/bibliography.scss
+++ b/app/assets/stylesheets/modules/bibliography.scss
@@ -1,5 +1,25 @@
-.show-document {
+.bibliography-contents {
+  margin-left: 0;
+  padding-left: 0;
+
   h3 {
     color: $brand-primary;
+    margin-left: 0;
+    padding-left: 0;
+  }
+
+  .bibliography-list {
+    margin-left: 0;
+    padding-left: 0;
+
+    a {
+      white-space: nowrap;
+    }
+
+    .bibliography-body {
+      margin-left: 2em;
+      max-width: 80ch;
+      text-indent: -2em;
+    }
   }
 }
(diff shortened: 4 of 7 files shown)


app/assets/javascripts/bibliography.js
```diff
@@ -0,0 +1,46 @@
...
+    init: function (el) {
+      var $el = $(el);
+      var data = $el.data();
+      // Add a placeholder so flashes of text are not as significant
+      var placeholder = '<div class="bibliography-placeholder">' +
+                          '<h3 class="col-md-9">Bibliography</h3>' +
+                          '<div class="col-md-9 bibliography-list"></div>'+
+                        '</div>';
+      $el.html(placeholder);
+      $.ajax({
```
mejackreed: Since we are just using the JSON api, `$.getJSON` is a convenience for jQuery.

app/assets/javascripts/bibliography.js
```diff
@@ -0,0 +1,47 @@
...
+          'f[format_main_ssim][]': 'Reference',
+          format: 'json',
+          rows: '1000'
+        }
+      }).done(function (response) {
+        var baseUrl = 'http://zotero.org/groups/1051392/items/';
+        for (var i = 0; i < response.response.docs.length; i++) {
+          var bibEntry = response.response.docs[i];
+          var html = '<p class="bibliography-body">' +
+                      bibEntry.formatted_bibliography_ts +
```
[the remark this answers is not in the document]
drh-stanford: This is our Solr field name.

app/assets/javascripts/bibliography.js
```diff
@@ -0,0 +1,46 @@
...
+  };
+
+  global.Bibliography = Bibliography;
+}(this));
+
+Blacklight.onLoad(function () {
+  'use strict';
+
+  $('.bibliography-contents').each(function (i, element) {
+    Bibliography.init(element); // eslint-disable-line no-undef
```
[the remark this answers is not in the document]
drh-stanford: It's defined in the global context, so I'm not sure what it's complaining about.

app/assets/javascripts/bibliography.js
```diff
@@ -0,0 +1,46 @@
...
+          $el.find('.bibliography-list').append(html);
+        }
+      });
+    }
+  };
+
+  global.Bibliography = Bibliography;
+}(this));
+
+Blacklight.onLoad(function () {
```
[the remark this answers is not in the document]
drh-stanford: It's defined in the global context, so I'm not sure what it's complaining about.

app/assets/javascripts/bibliography.js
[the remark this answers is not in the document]
jkeck: You can add

```
/* global Bibliography */
```

at the top of the file I believe that'll stop these.

app/views/catalog/_bibliography_default.html.erb
```diff
@@ -0,0 +1,5 @@
+<div class='col-md-12 bibliography-contents'
+     data-path="<%= search_exhibit_catalog_path(exhibit_id: @exhibit) -%>"
+     data-parentid="<%= @document.id -%>"
```
jkeck: Kinda pedantic but the `-` at the end of the ERB isn't really necessary here.

app/views/catalog/_bibliography_default.html.erb
```diff
@@ -0,0 +1,5 @@
...
+     data-sort="author_sort asc, pub_year_isi asc, title_sort asc">
```
jkeck: I wonder if we can derive this value from the `blacklight_config`somehow, or at the very least throw this into a configuration so this isn't hard-coded into the partial.

Just commenting on the potential fragility, we don't necessarily need to do anything about it now.

app/assets/javascripts/bibliography.js
```diff
@@ -0,0 +1,45 @@
...
+      var $el = $(el);
+      var data = $el.data();
+      // Add a placeholder so flashes of text are not as significant
+      var placeholder = '<div class="bibliography-placeholder">' +
+                          '<h3 class="col-md-9">Bibliography</h3>' +
+                          '<div class="col-md-9 bibliography-list"></div>'+
+                        '</div>';
+      $el.html(placeholder);
+      $.getJSON(data.path, {
+          'f[related_document_id_ssim][]': data.parentid,
```
jkeck: It looks like this is the only place where we use this `parentid`.  Would it make sense for us to pass this info in via the `path` attribute and not have to construct the `facet` params here?

jkeck:
Also, I just added #665 if you want to build on-top of that for your javascript testing.

app/assets/javascripts/bibliography.js
drh-stanford: I'm not sure I follow 100%. Is the idea to keep the facet params out of the JavaScript? I could construct the entire URL in the ERB view template if that's a better approach.

jkeck: I guess it just simplifies what's needed (html attribute wise) to instantiate the JS and then is/can be less tied to the implementation of "construct URL with these attributes" and could be more generically used to "replace contents of element X with response from request Y".

Totally could be me thinking about this way too abstractly though, so feel free to ignore. 😸  This is good work. 👍

drh-stanford:
Ok, I believe I've addressed all the comments, and this PR is ready for review now.

PR #661, opened 2017-10-03
update changelog

mejackreed:


CHANGELOG.md
```diff
@@ -19,14 +19,36 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
...
+
+### Changed
+- Indexes an author for a BibTeX file #628
+- Enhances BibTeX indexing to add parent druids #626
+- Enhances BibTeX indexing to skip records without a title #637
+- Enhances BibTeX indexing to skip records without keywords #644
+- Allows a user to upload a BibTeX file to index #639
+- Improve BibTeX indexing when BibTeX.parse lexer throws warnings #646
+- Improve BibTeX indexing by caching the title #645
+- Updates BibTeX fixtures to use all available Zotero fields #656
```
camillevilla: slight adjustment to this: I would say something like a "minimum set of Zotero fields". We only used a selection BibTeX fields relevant to Parker. Full compatibility with Zotero export will require some additional work.

mejackreed: Thanks for the suggestion! I've updated the PR to use the suggested text.

PR #666, opened 2017-10-04
generalizes the id creator allowing other Zotero records (or somethin…

### LINKED ISSUE #638: Generalize bibtex_config id creator
Right now it is specific to parker and zotero, this should be generalized:

https://github.com/sul-dlss/exhibits/blob/master/lib/traject/bibtex_config.rb#L11

mejackreed:
…g else)

Closes #638 

I'm not sure if there is a better way to do this until we have some more information about other uses of BibTeX upload.


jkeck:
#666 🤘 😈 🤘

mejackreed:
Thanks for the suggestion @drh-stanford . I have added a new field `bibtex_key_ss` that stores the key coming from the BibTeX data.

PR #668, opened 2017-10-04
Enables customize manifest url for a given exhibit

### LINKED ISSUE #648: Add a customization to implement a url scheme relacement in viewer admin page
Actionable ticket split out of #612 

In the viewer admin page create a url scheme form that can be used (per exhibit)

mejackreed:
Closes #648 

custommanifest

### DIFF
--- a/app/models/concerns/manifest_concern.rb
+++ b/app/models/concerns/manifest_concern.rb
@@ -5,4 +5,9 @@ module ManifestConcern
   def manifest
     fetch('iiif_manifest_url_ssi', nil)
   end
+
+  def exhibit_specific_manifest(custom_manifest_pattern)
+    return manifest if custom_manifest_pattern.blank?
+    custom_manifest_pattern.gsub('{id}', first('id'))
+  end
 end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20170821165811) do
+ActiveRecord::Schema.define(version: 20171004141834) do
 
   create_table "bibliography_services", force: :cascade do |t|
     t.string   "header"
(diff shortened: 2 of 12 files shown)


config/environments/development.rb
```diff
@@ -11,7 +11,7 @@
 
   # Show full error reports.
   config.consider_all_requests_local = true
-
+config.active_job.queue_adapter     = :sidekiq
```
jkeck: Accidentally left in?

mejackreed: Yikes.. yep

mejackreed: Fixed it!

app/models/concerns/manifest_concern.rb
```diff
@@ -5,4 +5,9 @@ module ManifestConcern
   def manifest
     fetch('iiif_manifest_url_ssi', nil)
   end
+
+  def exhibit_specific_manifest(custom_manifest_pattern)
+    return manifest if custom_manifest_pattern.blank?
+    custom_manifest_pattern.gsub('{id}', fetch('id', nil))
```
jkeck: I'm totally 👍 on this approach.  Just wanted to bring up an alternative for consideration (I don't think we know enough to know what is the right approach, but if this makes its way to Spotlight it might be worth thinking through).

If we use the `%{id}` syntax then we can use `String%` (e.g. `custom_manifest_pattern % { id: fetch('id', nil) }`).

The benefit of this being that we could allow an application to configure the set of document keys they will allow to be interpolated into a manifest.

mejackreed: Thanks for that feedback. Yeah I was also thinking that any Solr field could be used in the pattern too, but didn't want to go that far into it until we needed to.

Great suggestion. 🆗 to keep this approach for this pr?

drh-stanford: You should be able to use `first('id')` rather than `fetch('id', nil)`. See https://github.com/projectblacklight/blacklight/blob/dee8d794125306ec8d4ab834a6a45bcf9671c791/app/models/concerns/blacklight/document.rb#L89

db/schema.rb
```diff
@@ -399,8 +399,9 @@
   create_table "viewers", force: :cascade do |t|
     t.string   "viewer_type"
     t.integer  "exhibit_id"
-    t.datetime "created_at",  null: false
-    t.datetime "updated_at",  null: false
+    t.datetime "created_at",              null: false
+    t.datetime "updated_at",              null: false
+    t.string   "custom_manifest_pattern"
```
drh-stanford: i think `string` defaults to a limit of 255 characters. Is that enough for a IIIF URL?

app/models/concerns/manifest_concern.rb
mejackreed: Nice catch, I've updated to use your suggestion.

db/schema.rb
mejackreed: Thanks for catching this. I've updated to use `:text`

PR #670, opened 2017-10-04
add View on Zotero site to bibliography show page

### LINKED ISSUE #539: For a bibliography item enable link to Zotero bibliographic item

drh-stanford:
This PR closes #539. It adds a new partial that'll allow us to add blue buttons before the Edit button on bibliography show pages.

### After

screen shot 2017-10-05 at 12 11 34 pm

Also the change to `_bibliography_default.html.erb` removes the (not-needed) Bibliography section from the bibliography show page.

### DIFF
--- a/app/views/catalog/_bibliography_buttons_default.html.erb
+++ b/app/views/catalog/_bibliography_buttons_default.html.erb
@@ -1,0 +1,5 @@
+<% if @document.zotero_url.present? %>
+  <a href="<%= @document.zotero_url %>" class="btn btn-info btn-view-on-zotero pull-right">
+    View on Zotero site
+  </a>
+<% end %>
--- a/app/views/catalog/_bibliography_default.html.erb
+++ b/app/views/catalog/_bibliography_default.html.erb
@@ -1,9 +1,11 @@
-<div class='col-md-12 bibliography-contents'
-     data-path="<%= search_exhibit_catalog_path(exhibit_id: @exhibit) %>"
-     data-parentid="<%= @document.id %>"
-     data-baseurl="http://zotero.org/groups/1051392/items/"
-     data-sort="author_sort asc, pub_year_isi asc, title_sort asc">
-     <%# TODO: retrieve this sort configuration from the blacklight config %>
-     <h3 class="col-md-9">Bibliography</h3>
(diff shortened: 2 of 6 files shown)


drh-stanford:
I've added some logic that omits the View on Zotero site button if the link is just a regular BibTeX key.

app/views/catalog/_bibliography_buttons_default.html.erb
```diff
@@ -0,0 +1,12 @@
+<% if @document.reference? %>
+  <% begin %>
+    <% url = URI.parse(@document.bibliography_url) %>
```
mejackreed: This is great! Have you considered pulling this logic out of the view into a class? Perhaps that maybe useful as an extension point for other ways a BibTeX key could be expressed in a view?

drh-stanford:
Ok, I've moved that logic into the bibliography concern of SolrDocument.

app/views/catalog/_bibliography_default.html.erb
```diff
@@ -1,9 +1,11 @@
-<div class='col-md-12 bibliography-contents'
-     data-path="<%= search_exhibit_catalog_path(exhibit_id: @exhibit) %>"
-     data-parentid="<%= @document.id %>"
-     data-baseurl="http://zotero.org/groups/1051392/items/"
-     data-sort="author_sort asc, pub_year_isi asc, title_sort asc">
-     <%# TODO: retrieve this sort configuration from the blacklight config %>
-     <h3 class="col-md-9">Bibliography</h3>
-     <div class="col-md-9 bibliography-list"></div>
-</div>
+<% unless @document.reference? %>
```
drh-stanford: @aeschylus here's that conditional in the partial

PR #674, opened 2017-10-04
Add the ability to define exhibit based feature flags

### LINKED ISSUE #667: Add the ability to feature flag based on an exhibit
Our current feature flag ability is at the global exhibits level, if we can add exhibit level flags too, this could help us release functionality to specific exhibits sooner.

jkeck:
Closes #667

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -15,4 +15,9 @@ class ApplicationController < ActionController::Base
   # Prevent CSRF attacks by raising an exception.
   # For APIs, you may want to use :null_session instead.
   protect_from_forgery with: :exception
+
+  def feature_flags
+    @feature_flags ||= FeatureFlags.for(current_exhibit)
+  end
+  helper_method :feature_flags
 end
--- a/app/models/feature_flags.rb
+++ b/app/models/feature_flags.rb
@@ -1,0 +1,73 @@
+# frozen_string_literal: true
+
+##
+# A ruby class to allow for feature flags to be set given an exhibit context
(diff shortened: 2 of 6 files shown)


jkeck:
Was thinking about pulling this logic that knows how to take a given context (passed through `for`) and return context specific flags to be merged into the base so that this class didn't have to have knowledge of how to get exhibit specific settings.  Happy to follow that through if folks think it's a worthwhile abstraction.

jkeck:
From what I understand a UI for this is out of scope at the moment and I don't believe we have designs.

It seems like creating an active record backed feature flag implementation without a UI to update it would end up being more cumbersome to turn features on/off, or would require tooling to manage that state based on something like settings in the first place.

This implementation could allow for the same or similar API to be carried over once we do have an AR backed implementation.

ggeisler:
Yeah, I was just raising the question of: do we want a UI for feature flags? If the answer is yes, then I will work on a design in a new ticket.

mejackreed:
I was thinking about this. How would a UI for feature flags differ from the admin dashboard? They fill similar enough to me when I was thinking about it (but I may be wrong). I feel like we should trust administrative and curatorial users enough to show them new features as long as we give them proper context and information.

ggeisler:
Yep, we could definitely add it to the Configuration section of the exhibit admin instead. I think we'd want @caaster to weigh in on whether we want to give the power to exhibit admins or whether we want super-admins to control it. I suggested the latter approach because I sort of assumed @caaster might be concerned about individual admins having this control, but I certainly could be wrong about that.

Issue #650, opened 2017-10-02
Add document references to bibliography show page


jkeck:
The bibliography show page needs to have a section of the documents/manuscripts that it was associated with (based on the identifiers supplied in the bibliographies keywords)

aeschylus:
Based on some slack conversation we have the beginnings of two approaches based on the existing bibliography indexing strategy.

1.) Use the blacklight config to inject a partial in the appropriate place. I don't really understand how this works, but in theory is seems simpler. It might result in a slow page load due to blocking the request on a server-side API call.

2.) Reproduce what @drh-stanford has done in #660, and call out to the JSON API from the client page. This is pretty simple as well, and Darren has already given us a good example.

Long-term, it seems odd to call out to any API when the produced data is relatively static. Is there any way all these reference calculations can be done up front and stored until they need to be updated (once a year, say)?

PR #692, opened 2017-10-06
Add cited documents to bibliography show page

aeschylus:
is connected to #650.

### DIFF
--- a/app/assets/javascripts/cited_documents.js
+++ b/app/assets/javascripts/cited_documents.js
@@ -1,0 +1,40 @@
+/* eslint-disable camelcase */
+/* global CitedDocuments */
+/* global Blacklight */
+
+(function (global) {
+  var CitedDocuments;
+
+  CitedDocuments = {
+    init: function (el) {
+      var $el = $(el);
+      var data = $el.data();
+      var solrQueryString = data.documentids.join(' OR ');
+
+      $.getJSON(data.path, {
+        q: solrQueryString,
+        format: 'json',
+        rows: 1000
+      }, function (response) {
+        response.response.docs.forEach(function(citedDocEntry) {
+          var html = '<li class="cited-documents-body">' +
+                ' <a href="' + citedDocEntry.id + '">' +
+                citedDocEntry.title_full_display +
+                '</a>' +
+                '</li>';
+          $el.find('.cited-documents-list').append(html);
+        });
+      });
+    }
+  };
+
+  global.CitedDocuments = CitedDocuments;
+}(this));
+
+Blacklight.onLoad(function () {
+  'use strict';
+
+  $('.cited-documents-contents').each(function (i, element) {
+    CitedDocuments.init(element); // eslint-disable-line no-undef
+  });
+});
--- a/app/assets/stylesheets/modules/bibliography.scss
+++ b/app/assets/stylesheets/modules/bibliography.scss
@@ -1,4 +1,5 @@
-.bibliography-contents {
+.bibliography-contents,
+.cited-documents-contents {
   margin-left: 0;
   padding-left: 0;
 
@@ -8,7 +9,8 @@
     padding-left: 0;
   }
 
-  .bibliography-list {
+  .bibliography-list,
+  .cited-documents-list {
     margin-left: 0;
     padding-left: 0;
 
@@ -20,6 +22,11 @@
       margin-left: 2em;
       max-width: 80ch;
       text-indent: -2em;
+    }
+
+    .cited-documents-body {
+      margin-left: 2em;
+      max-width: 80ch;
     }
   }
 }
--- a/spec/features/manuscript_display_bibliography_spec.rb
+++ b/spec/features/manuscript_display_bibliography_spec.rb
@@ -1,0 +1,54 @@
+# coding: utf-8
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.feature 'Cited manuscripts display on the bibliography show page', type: :feature do
+  let(:exhibit) { FactoryGirl.create(:exhibit, slug: 'default-exhibit') }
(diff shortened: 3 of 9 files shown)


app/assets/javascripts/cited_documents.js
```diff
@@ -0,0 +1,44 @@
...
+
+(function (global) {
+  var CitedDocuments;
+
+  CitedDocuments = {
+    init: function (el) {
+      var $el = $(el);
+      var data = $el.data();
+      console.log(data);
+      $.getJSON(data.path, {
```
drh-stanford: @aeschylus I'm not sure what's the best way to do this but for each related document_id you can use the show route for it to get the title via JSON like so:

  https://exhibits.stanford.edu/parker-2/catalog/wp146tq7625.json

I don't know how to construct the Blacklight JSON API query that allows you to put in multiple document_id's, because this:

  https://exhibits.stanford.edu/parker-2/catalog.json?f[id][]=wp146tq7625&f[id][]=abcd

doesn't seem to work.

drh-stanford:
@aeschylus For the specs, to get a bibliography entry, you can do the same thing that the bibliography_resource_integration_spec does. Namely, create a `BibliographyResource` object and then use it for a `to_solr` call. That'll give you a hash that will have the related manuscripts data.

See the `before`/`after` and `let`'s in this bit of code:

https://github.com/sul-dlss/exhibits/blob/master/spec/features/bibliography_resource_integration_spec.rb#L6-L30

app/assets/javascripts/cited_documents.js
```diff
@@ -0,0 +1,39 @@
...
+
+  CitedDocuments = {
+    init: function (el) {
+      var $el = $(el);
+      var data = $el.data();
+      data.documentids.forEach(function(documentId) {
+        $.getJSON(data.path, {
+          'f[id][]': documentId,
+          format: 'json',
+          rows: '1000'
```
drh-stanford: Since you already have the specific document ID, you can change the `rows` to 1.

app/assets/javascripts/cited_documents.js
```diff
@@ -0,0 +1,39 @@
...
+/* global Blacklight */
+
+(function (global) {
+  var CitedDocuments;
+
+  CitedDocuments = {
+    init: function (el) {
+      var $el = $(el);
+      var data = $el.data();
+      data.documentids.forEach(function(documentId) {
```
mejackreed: This could be potentially problematic, especially with a reference that cites all Parker items (500+ HTTP calls). I think we could get a away with one query like this:
```
$.getJSON(data.path, {
  q: (sp968bx9690 OR zy353np1659... ),
  format: 'json',
  rows: 1000
}, function (response) {
```

spec/features/manuscript_display_bibliography_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+    # render the resource show page
+    visit spotlight.exhibit_solr_document_path(exhibit_id: exhibit.slug, id: resource_id)
+  end
+
+  after :all do
+    ActiveJob::Base.queue_adapter = :test # restore
+  end
+
+  scenario 'cited documents element data required by async loader comes through to front end' do
+    puts page.body
```
mejackreed: Can we remove this logging statement?

spec/features/manuscript_display_bibliography_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+    expect(page).to have_css('div.cited-documents-contents[data-path="/default-exhibit/catalog"]')
+    expect(page).to have_css("div.cited-documents-contents[data-parentid=\"#{resource_id}\"]")
+    expect(page).to have_css('div.cited-documents-contents[data-documentids]')
+    expect(page.find('div.cited-documents-contents[data-documentids]')['data-documentids']).to eq(citationsString)
+    puts page.find('div.cited-documents-contents[data-documentids]')['data-documentids']
```
mejackreed: Can we remove this logging statement?

db/schema.rb
```diff
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20171005231502) do
+ActiveRecord::Schema.define(version: 20171017223044) do
```
drh-stanford: you can omit this change

spec/features/manuscript_display_bibliography_spec.rb
drh-stanford: you can remove this debugging code

app/assets/javascripts/cited_documents.js
```diff
@@ -0,0 +1,42 @@
...
+(function (global) {
+  var CitedDocuments;
+
+  CitedDocuments = {
+    init: function (el) {
+      var $el = $(el);
+      var data = $el.data();
+      var solrQueryString = data.documentids.reduce(function(query, nextId, index, array){
+        return (index < array.length - 1) ? query + nextId + ' OR ' : query + nextId;
+      }, '');
```
drh-stanford: is there a `.join` method for JavaScript arrays?

aeschylus: Ha, yes. Not sure why reduce occurred to me first.

app/assets/stylesheets/modules/bibliography.scss
```diff
@@ -8,18 +9,19 @@
...
     a {
       white-space: nowrap;
     }
 
-    .bibliography-body {
+    .bibliography-body,
+    .cited-documents-body {
       margin-left: 2em;
       max-width: 80ch;
-      text-indent: -2em;
```
mejackreed: I think this might be causing an issue on the bibliography view:

screen shot 2017-10-27 at 7 41 52 pm

Could we modify the selector or classes here?

Issue #651, opened 2017-10-02
Implement bibliographic items in search results


mejackreed:
Described and designed in #512 & #575

From the design document:
 - [x] 1: Confirm / test reference facetability
 - [x] 2: Add resource type thumb image/placeholder

![designdoc](https://user-images.githubusercontent.com/101482/31030936-f0a759ba-a50b-11e7-8599-0e1f45642968.png)

camillevilla:
Will depend on analysis in #649

drh-stanford:
For the sorting algorithm in #632, we'd like to add Author-Date-Title, so we need to index the year into the `pub_year_isi` field, and in CatalogController add:

```
config.add_sort_field 'author_sort asc, pub_year_isi asc, title_sort asc', label: 'author-date'
```

jkeck:
WRT faceting, we do get the appropriate resource type for the Bibliography references themselves, however; ensuring that the other parker objects get a Resource Type of "Manuscript" may be the tricker part.  Digging into that now.

jkeck:
@blalbrit it appears that the Resource Type value that we'll want to be present in the facet is `Archive/Manuscript`.  You can see all the Resource Types available to us in the [screenshot below](#sw-screenshot).

The indexing code in [stanford-mods](https://github.com/sul-dlss/stanford-mods/blob/ccde41b874dfcea4bda627df883adcbb4c95da96/lib/stanford-mods/searchworks.rb#L313-L314) indicates that the `typeOfResource` needs to be `mixed material` (e.g. `<typeOfResource manuscript="yes">mixed material</typeOfResource>`).  I'm not sure what the `manuscript="yes"` bit is about, but that's what I see in the wild for documents that have this Resource Type.

<img name="sw-screenshot" width="389" alt="sw-resource-tpes" src="https://user-images.githubusercontent.com/96776/31291350-7189fd9e-aa84-11e7-8a71-00105ce71abe.png">

blalbrit:
Cool @jkeck - so just to confirm, typeOfResource should look like this example: https://purl.stanford.edu/cf386wt1778.mods  ?

jkeck:
Exactly @blalbrit.

jkeck:
#698 has been submitted as a potential solution to `Add resource type thumb image/placeholder`.  If that is acceptable and we now know the data requirements for `Confirm / test reference facetability`, would this ticket be closable?

mejackreed:
Closing this with #698 being 🚢

PR #698, opened 2017-10-06
Override blacklight's thumbnail methods to provide a default for bibliography items

jkeck:
Connects to #651 

## List View
list-view

## Admin View
admin-view

## Masonry View
masonry-view

## Gallery View
gallery-view

I didn't add the Slideshow view because @blalbrit said we could disable that view.


mejackreed:
Not a review, but just a comment. I like the approach of #2 better as it feels "more right". But the simplicity of approach #1 is hard to argue with.

Have you looked into which methods need to be overridden for approach #2? Does that live in Spotlight or somewhere else?

jkeck:
It's not so much overridden, it is just that we'll need to configure `thumbnail_method` in the view configuration.  That method will effectively be something along the lines of:

```ruby
def index_thumbnail_method(*args)
  if document.reference?
    image_tag('fallback-image')
  else
    blacklight_render_method(some_subset_of_args)
  end
end
```

Also, because of the different possible thumbnails used in different view types we would either need to provide different methods for different views or provide logic w/i the method to point to the appropriate sized thumbnail.

lib/traject/bibtex_config.rb
```diff
@@ -116,6 +116,18 @@
...
+}
+
+to_field 'thumbnail_url_ssm', lambda { |_, accumulator, _|
+  accumulator << ActionController::Base.helpers.asset_path('default-thumbnail-book.png')
+}
+
+to_field 'large_image_url_ssm', lambda { |_, accumulator, _|
+  accumulator << ActionController::Base.helpers.asset_path('default-thumbnail-book.png')
+}
+
```
drh-stanford: I think you can do these with the `literal` macro, namely:

```
to_field 'large_image_url_ssm', literal(ActionController::Base.helpers.asset_path('default-thumbnail-book.png').to_s)
```

mejackreed:
Gotcha thanks @jkeck for that clarification.

Perhaps with your comment in mind, I'm more inclined to lean towards option #2. For the following reasons:
 - It doesn't force us to index an asset path for display purposes only (this seems suboptimal)
 - It allows us to tap into (perhaps in the correct seam?) the ability for falling back when there is no indexed thumb. This seems like it could be desirable.
 - Would potentially force a reindex in production for an unknown exhibit just to get access to a new icon/image.

What do you think?

PR #699, opened 2017-10-07
Update BibTeX upload UI

### LINKED ISSUE #695: Short-term fixes for UI: uploading bibliography file in Admin
See discussion in design ticket #586
These are small changes to the `Curation > Items > Add Items` area. A more comprehensive redesign will require breaking changes, and possibly upstream changes in Spotlight. 

@ggeisler:
If we decide to go with something simpler (hopefully just for the short-term) to avoid breaking Spotlight, I suggest some minor updates to the existing implementation:

- [x] Change the label of the bib button to "BibTex"

- [x] Use horizontal alignment of the label and input (for consistency with the presentation with the other resource options)

- [x] Add the following hint text below the file picker input (assuming this wording makes sense to others):
> Each BibTex file entry will be added as an exhibit item with a “Reference” resource type.

## Current UI
curation admin

camillevilla:
Closes #695 

This PR updates our BibTeX file upload UI:
- shortens button text
- uses a horizontal label + form to match other upload options
- adds helper text

## Before
bibtex_upload_form_before

## After
bibtex_upload_form_after

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ### Added
 ### Changed
+- Updates BibTeX file upload UI #699
 ### Deprecated
 ### Removed
 ### Fixed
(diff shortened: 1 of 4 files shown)


ggeisler:
Looks great (not reviewing code, just the UI in the screenshot). One super minor thing that is my fault because I forgot to fix it in the mockup: in the hint text below the input, the "x" in BibTex should be capitalized.

camillevilla:
ah,good catch, @ggeisler! I've updated the text and screenshot.

Issue #671, opened 2017-10-04
Refactor indexing pipeline to use Bibliography model for information extraction


drh-stanford:
The `Bibliography` model has some logic that cleans up data and renders HTML references, etc. We need a way to use the `Bibliography.new(record.to_s)` pattern to load the data into the model during the indexing. Otherwise, we won't get the logic in the model.

See https://github.com/sul-dlss/exhibits/blob/master/lib/traject/bibtex_config.rb#L114-L121 for an example of how we use the Bibliography model in the pipeline.

cmh2166:
Perhaps this is an ideal pairing or tech team discussion opportunity? 

A proposal: 
1. we can first go through some options in traject+ (traject plus our additions from exhibits or DLME) in refactoring the `Bibliographic` model, like the use of macros; 
2. get work started on refactoring as decided for that model in exhibits; 
3. then maybe at least mention or foreshadow other exhibits models, representations, and data sources we currently have or will have that could leverage this work (if not already done somewhere).

I'd prefer if possible this happen when @cbeer could join as well, perhaps early next Monday if he's not available before.

drh-stanford:
@cmh2166 has a feature branch that has several ideas for refactoring here -- https://github.com/sul-dlss/exhibits/blob/671-refactor-traject-cmh2166. Reopening this issue

PR #706, opened 2017-10-09
refactor BibTeX extraction macros

drh-stanford:
This PR is connected to #671 and it does a few things:

1. Migrates all extraction-related code from bib_config into BibTeX macros
2. Reuses some common macros used by DLME (`Macros::General` and `Extraction`)
3. Fixes bug where `\textbackslash` wasn't being properly removed during parsing
4. Consolidates application logic into `Bibliography` model and uses it for `BibReader`

Some benchmarks from my laptop instance:
```
2017-10-09T15:45:36-07:00  INFO Reading from /Users/drh/Local/ParkerProductionBibliography.bib
2017-10-09T15:45:40-07:00  INFO    Indexer with 5 processing threads, reader: BibReader and writer: Traject::DebugWriter
2017-10-09T15:51:10-07:00  INFO finished Indexer#process: 6165 records in 333.451 seconds; 18.5 records/second overall.
```

cc: @cmh2166

### DIFF
--- a/lib/traject/bibtex_config.rb
+++ b/lib/traject/bibtex_config.rb
@@ -2,9 +2,12 @@
 
 require_relative 'bib_reader'
 require_relative 'macros/bibtex'
+require_relative 'macros/general'
+
 require 'bibliography'
 require 'active_support/core_ext/object/blank'
 extend Macros::BibTeX
+extend Macros::General
 
 settings do
   provide 'reader_class_name', 'BibReader'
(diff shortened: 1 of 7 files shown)


lib/traject/bibtex_config.rb
```diff
@@ -17,71 +20,26 @@
...
-
-# Druids are kept as tags (keywords) in the BibTeX::Entry
-to_field 'related_document_id_ssim', lambda { |record, accumulator, _context|
-  record.keywords.to_s.split(',').map(&:strip).each do |druid|
-    accumulator << druid if druid =~ Exhibits::Application.config.druid_regex
-  end
-}
+to_field 'bibtex_ts', extract_bibtex_raw
+to_field 'formatted_bibliography_ts', extract_bibtex_formatted_bibliography
+to_field 'related_document_id_ssim', extract_bibtex_related_document_ids
```
cmh2166: given rec below on adding traject options support to extract_bibtext_field(), this line would become: 

```ruby
to_field 'related_document_id_ssim', extract_bibtex_field(:keywords, split: ',', trim: true, match: Exhibits::Application.config.druid_regex)
``` 


which isn't as pretty in the config, but makes it easier to call on split, trim, match Extraction methods on a varying field mapping basis. This is something @camillevilla could use for the `annote` break later, for example.

lib/traject/bibtex_config.rb
```diff
@@ -17,71 +20,26 @@
...
+to_field 'location_ssi', extract_bibtex_field(:address)
+to_field 'university_ssim', extract_bibtex_field(:school)
+to_field 'edition_ssm', extract_bibtex_field(:edition)
+to_field 'series_ssi', extract_bibtex_field(:series)
+to_field 'thesis_type_ssm', extract_bibtex_field(:type)
+to_field 'volume_ssm', extract_bibtex_field(:volume)
+to_field 'issue_ssm', extract_bibtex_field(:issue)
+to_field 'pages_ssm', extract_bibtex_field(:pages)
+to_field 'doi_ssim', extract_bibtex_field(:doi)
+to_field 'general_notes_ssim', extract_bibtex_field(:annote)
```
cmh2166: Same here as with the changes requested below - once extraction macro support is setup for `extract_bibtext_field`, and annotations are concatenated following whatever character, this can become: 

```
to_field 'general_notes_ssim', extract_bibtex('annote', split: '|', trim: true)
```

discussed briefly with @camillevilla

lib/traject/macros/bibtex.rb
```diff
@@ -9,28 +11,77 @@ module BibTeX
...
+        html = Bibliography.new(record.to_s).to_html
+        doc = Nokogiri::HTML(html)
+        li = doc.at_css('ol li')
+        reference = li.children.to_html if li.present? # extract just the reference from <li>
+        accumulator << reference.to_s
+      end
+    end
+
+    # Druids are kept as tags (keywords) in the BibTeX::Entry
+    def extract_bibtex_related_document_ids
```
cmh2166: if the extraction options are taken into play, this method is no longer needed.

lib/traject/macros/extraction.rb
```diff
@@ -0,0 +1,71 @@
...
+          v.split(splitter)
+        end
+      end
+
+      def transform(values)
+        options.inject(values) { |memo, (step, params)| public_send(step, memo, params) }
+      end
+
+      def translation_map(values, maps)
+        translation_map = Traject::TranslationMap.new(*Array(maps))
```
cmh2166: this is a file for later discussion comment: i think some of the possible metadata tickets (at least for parker as parker is using traject) could benefit from lookups / translation_maps.

lib/traject/macros/bibtex.rb
```diff
@@ -9,28 +11,77 @@ module BibTeX
...
         when :ref_type # use record.type to get BibTeX type
           accumulator << BIBTEX_ZOTERO_MAPPING[record.type] if BIBTEX_ZOTERO_MAPPING.include?(record.type)
         when :year # built-in method in BibTeX::Entry
           accumulator << record.year
-        when :raw # serialization into BibTeX format
-          accumulator << record.to_s
         else # use the dynamically constructed BibTeX method to fetch the field rather than fetch/get
-          accumulator << record.send(key).to_s.presence if record.respond_to?(key)
+          if record.respond_to?(key.to_sym)
+            result = record.send(key.to_sym).to_s.presence
```
cmh2166: below this line, add the following to enable extraction macros to run on the values array:

```
result = Macros::Extraction.apply_extraction_options(result, options)
```

This allows at least the cases that are _not_ handled in lines 71-77 to support the extraction additions. I think a general pattern to highlight with future traject use is to move these sorts of field specific handlers to separate methods to make it really clear they aren't available for similar macros usage.

lib/traject/bibtex_config.rb
drh-stanford: i moved that code into the bibtex macros

lib/traject/bibtex_config.rb
drh-stanford: ok, i'm keeping this little bit of extraction logic in bib_config for now.

lib/traject/macros/extraction.rb
drh-stanford: yes our `:ref_type` uses a translation map that we could use this for

drh-stanford:
@cmh2166 I've made the changes you've requested and I've forced pushed them to this PR. Thanks!

lib/traject/macros/extraction.rb
cmh2166: i can add that set up as a new PR then. thanks!

lib/traject/macros/bibtex.rb
cmh2166: thank you!

lib/traject/macros/extraction.rb
camillevilla: 👍

camillevilla:
Er, @drh-stanford I forgot to ask: can you add something to the CHANGELOG?

Issue #679, opened 2017-10-05
Design for IIIF badges on show page and results page


mejackreed:

ggeisler:
This is a feature that is trickier than it seems at first glance. I don't think we can simply add the IIIF icon alone and leave it at that, because relatively few users will know what IIIF is, and even fewer understand that they can drag the icon to a IIIF viewer. At the same time, we don't want to add a lot of text around the icon to explain what it's for, especially on the results page where it'll appear once for every search result. For example, I have a bunch of alternatives similar to this:

<img width="881" alt="spotlight_graffle__parker__public___search_results_with_iiif_badge 2" src="https://user-images.githubusercontent.com/101482/31300619-83b3523e-aaa9-11e7-9f84-4072538cf876.png">

where the "what's this" link expands to show more info:

----
<img width="558" alt="more-info" src="https://user-images.githubusercontent.com/101482/31300670-c2dda536-aaa9-11e7-82b0-66fb37f74bea.png">

----

but all the "what's this" links are going to be very repetitive on a results page.

Alternatively, how about we create a new icon that enables us to remove the text but still provides an indication that the IIIF icon has an action associated with it, like this:

<img width="1232" alt="results-icon-only" src="https://user-images.githubusercontent.com/101482/31300718-03d32142-aaaa-11e7-956b-9a3b4185c9ee.png">

This is a little less repetitive (without the text). But there's still the problem of some users not knowing what it does. So for users who do know, they can drag it to a IIIF viewer. For curious users who don't know, they'll likely click it. Unfortunately, clicking will go to the underlying linked resource, which is a IIIF manifest and not very informative for a naive user. But if we use Javascript to intercept the click event and rewrite the `href` value, we can make the link go to this page instead: https://library.stanford.edu/projects/international-image-interoperability-framework/viewers

That way the user can learn about what IIIF viewers are, etc.

If this approach seems feasible to others, we'll need to create a new icon and check with @snydman about the status of the lIIF viewers page we want to link to (the link to it is hidden in the sidebar of the [main IIIF page](http://library.stanford.edu/projects/international-image-interoperability-framework) on the Library site, so JV and I are not sure whether it is still a draft or needs updating, etc.).

If this approach seems good to others, the show page will use the same icon and click behavior:

<img width="1207" alt="show-page-icon-only" src="https://user-images.githubusercontent.com/101482/31300971-86e695ae-aaab-11e7-8cef-35584f5c4251.png">

[...]

PR #707, opened 2017-10-10
IIIF drag and drop added to Mirador show page

### LINKED ISSUE #680: Add IIIF badge to show page for objects w/ a manifest
Some information about IIIF drag and drop here:

http://zimeon.github.io/iiif-dragndrop/

Design in #679

### LINKED ISSUE #693: Non-image items with a custom manifest URL still show a viewer
Introduced in #668 

We need some better way not to show a viewer when there is a custom pattern.

mejackreed:
Closes #680 

Also closes #693 

I'm not sure why we have to drag twice. Is this an issue w/ mirador? cc @aeschylus (I remember something like this from before)

dnd


app/helpers/application_helper.rb
```diff
@@ -13,4 +13,12 @@ def document_collection_title(value:, **)
   def document_leaflet_map(document:, **)
     render_document_partial(document, 'show_leaflet_map_wrapper')
   end
+
+  ##
+  # @param [String] manifest
+  def iiif_drag_n_drop(manifest)
+    link_to mirador_index_path(manifest: manifest), class: 'iiif-dnd pull-right', data: { turbolinks: false } do
```
jkeck: I _think_ we want the base of the link to go to an information page about IIIF on library.stanford.edu that was described in #679 (but maybe wasn't explicitly called out as acceptance criteria in the issue.

jkeck:
@mejackreed what's the manifest URL you're using that you're seeing the double drag-and-drop behavior?  I seem to remember that being something to do w/ a miss-match in the url and the `@id` in the manifest.

app/helpers/application_helper.rb
ggeisler: Yes, we want a click to go to this page:

https://library.stanford.edu/projects/international-image-interoperability-framework/viewers

We might want to update that page but I think it is good enough to serve the purpose as is.

mejackreed: Thanks for catching that. I've updated the PR to do that.

mejackreed:
@jkeck looks like its our indexing code that is indexing `manifest.json`.. creating an issue there

PR #712, opened 2017-10-10
use BibTeX prefix for Traject pipeline throughout

drh-stanford:
This PR just keeps our file/class naming consistent for the BibTeX Traject pipeline code.


camillevilla:
@drh-stanford I was surprised that Rails didn't get tripped up with the capitalization (`require_relative 'bibtex_reader` and `BibTeXReader`). I tried things out locally and things seem to be working just fine  ¯\\_(ツ)_/¯

Issue #697, opened 2017-10-06
Display repeated instances of the same field on separate lines


caaster:
Story: Using multiple instances of the same element to structure information indicates that each instance is a distinct unit. Concatenating these units into a single sentence regardless of their individual content or internal punctuation blurs these distinctions, and is more difficult for the user to parse. Exhibit creators would like to display each instance on a separate line for greater clarity.

Acceptance criteria:
Each metadata value indexed individually to a field should appear on a separate line in the item display.

mejackreed:
How can we delineate when something should delimited by a comma vs a separate line? This has potential implications on how the page displays and could have unwanted side effects. For example, if an item has many subjects, should these all be separate lines?

Do we have some concrete examples of metadata where/why this is needed?

I think there are a lot of design implications here and would defer to @ggeisler on how to handle this.

ggeisler:
This issue has been resolved for the case of repeated notes (#715). 

To deal with other fields, do we know which other fields are candidates for repeated instances? We have a couple of potential patterns in use that could be applied for repeated instances of the same field:

* The pattern used for notes, where there is one field label (e.g., "Notes") and the repeated values displayed in a `ul`. In this case, all values are displayed.

* The pattern used (in progress, I think) for the text title field (table of contents) where it is a single field but has many values, separated by a `-` (I think). In this case the plan is to put all the values in a expand/collapse structure, so that there is a single field label and the initial field value is "Show". When clicked, all values are displayed, each on its own line. The list can be long, but initially it isn't shown, and when shown it can be collapsed back to the "show" link.

I'm hoping we can stick to using one or the other of these patterns for other cases but I think we might need to be specific about the potential cases so we can decide which pattern might be most appropriate for each.

arcadiafalcone:
An overview of fields in MODS that may have multiple values.

More common:
1. Alternative title (cf. Parker)
2. Name/role - names per record, names per role, and roles per name
3. Genre
4. Date - different date types within a single originInfo element, and multiple originInfo elements with non-contiguous dates or different event types
5. Place of origin - multiple within a single originInfo element and multiple originInfo elements with different event types
6. Language
7. Physical description note
8. Note
9. Subject
10. Identifier
11. Related item

Less common but possible: translated title, type of resource, publisher, and form.

[...]

Issue #544, opened 2017-09-18
Index notes separately by display label - new MODS to SOLR mapping


caaster:
Story:
The displayLabel attribute is often used to create distinct categories of notes specific to the content, and the content of a category may function as a controlled vocabulary. Exhibit curators would like to leverage this enhanced granularity by distinguishing notes as separate, individually labeled fields for display and faceting, based on the specific displayLabel values provided in the metadata.

Acceptance criteria:
1. A separate display field should be available for each value of /mods/note/@displayLabel appearing in the exhibit metadata, labeled with that value and containing the value of the corresponding /mods/note.
2. A separate facet field should be available for each value of /mods/note/@displayLabel appearing in the exhibit metadata, labeled with that value and containing the value of the corresponding /mods/note.

arcadiafalcone:
This should be labeled data-contingent indexing rather than MODS to Solr mapping (my mistake).

blalbrit:
For Parker content - this object has multiple notes fields with arbitrary display labels: https://purl.stanford.edu/cf386wt1778.mods

jkeck:
I have some concerns about turning any note's `displayLabel` into a facet.  I worry that this could get particularly unwieldy, and since the `displayLabel` value seems like it can have different contexts in different records (e.g. "Height (mm)" vs "Height" vs. "Height (something else)") that we wouldn't necessarily want to all be disparate facets.  Also, this could also get really difficult to handle in the admin UI for metadata fields and facets.

Is there a `displayLabel` vocabulary that we could generate that lists the known `displayLabel` values that we would like to have this behavior and the rest would fall into a generic note field?

blalbrit:
I think a vocabulary that could drive facets would be good. For Parker, I'm more concerned about the display side of things than the faceting (that is - I specifically want my arbitrary displayLabels to show up, not very concerned about having them facetable - though I recognize there is a broader metadata story about the facets. Just want to clarify what's needed for Parker vs. bigger story.)

camillevilla:
@blalbrit I have a few questions for the Parker context:
We find notes with `displayLabel` under both `<relatedItem type="constituent">` and `<physicalDescription>`.

[...]

Issue #543, opened 2017-09-18
Index notes separately by display label, depending upon data


caaster:
Story:
Exhibit curators would like notes without a displayLabel attribute to be available for display as generic notes in the item-level view, even if they have a type attribute value.

Acceptance criteria:
The “Note” field in the item-level view should include all values of /mods/note without a displayLabel attribute.

arcadiafalcone:
It looks like bits of this got swapped with #544 -- the title and tags for this issue go with the description in that ticket. The title for this ticket should be "All notes without a displayLabel attribute should display in Notes field", and it should be tagged MODS to Solr mapping instead of data-contingent indexing.

PR #715, opened 2017-10-10
Support multiple notes per bibliography resource

### LINKED ISSUE #685: Analysis: Support multiple annotations per bibliography resource
Our BibTeX reader only captures a single annotation! e.g. [book.bib](https://github.com/sul-dlss/exhibits/blob/676-bib-show-page-annotations/spec/fixtures/bibliography/book.bib) has 3 `annote` values. When I checked our our BibTeX `record`, we only get the last annotation.

## Design
show page

design notes

#### BibTeX:
```
@book{http://zotero.org/groups/1051392/items/TXXUJDG2,
	address = {Coimbra},
	title = {A {Carta} ou {Memória} do {Cruzado} {Inglês} {R}. para {Osberto} de {Bawdsey} sobre a conquista de {Lisboa} em 1147},
        …
	annote = { KAW},
	annote = {CCC MS 470 [p. 9, and referred to (without MS number) throughout]reproductions from CCC MS 470, ff. 125-146, in the appendix},
	annote = {The following values have no corresponding Zotero field:custom1: 23}
}
```
#### BibTeX::Entry
```
#<BibTeX::Entry 
  address = Coimbra, 
  title = A Carta ou Memória do Cruzado Inglês R. para Osberto de Bawdsey sobre a conquista de Lisboa em 1147, 
  …
  annote = The following values have no corresponding Zotero field:custom1: 23
>
```

### Possible routes forward (please feel free to edit and add notes below)

1. mess with our BibTex translator so that we get something more like:
`annote = {KAW, The following values have no corresponding Zotero field:custom1: 23}}` and possibly make a PR to [zotero/translators](https://github.com/zotero/translators)) (@camillevilla)
2. adjust our Traject settings to deal with multiple `annote` values (in consultation with @cmh2166) 
3. issue a PR to bibtex-ruby to handle concatenation of multiple values with the same key (suggested by @drh-stanford)
4. [zotero-better-bibtex](https://github.com/retorquere/zotero-better-bibtex) (@mejackreed)

camillevilla:
Closes #685 
Design #605 

This PR:
- updates our Zotero --> BibTeX --> Traject --> Solr pipeline so that all notes are captured
- updates all fixtures with `annote` values (multiple notes are separated by `\textbar` (`|` in LaTeX land)
- adds paragraph separators to the notes (pending discussion below)

## Todo
- [x] [ Zotero --> BibTeX translator](https://github.com/camillevilla/bibtex-translator) that separates notes with `|` character
  - [ ] PR in zotero/translators
  - [x] Add translator to team Google drive
- [x] BibTeX --> Traject
- [x] update fixtures `annote` section
- [x] Blacklight config / helper for presenting notes on separate lines in show page (1: `<ul>` ideal, 2: `<p>` ok #605)
- [x] bib resource integration tests
- [x] unit test

## Questions:

- [x] How do existing exhibits use the `general_notes_ssim` field? Is this too much top-down control over how notes are displayed? 

- [x] How can I refactor this and make it more Blacklight-ish? 

## Before 
Before [6d38407](https://github.com/sul-dlss/exhibits/pull/715/commits/6d384075ab7c3a41968c05c05f8ce629a1d09fe0) we were only captured one note per bibliography resource
single_note

Before [18696d1](https://github.com/sul-dlss/exhibits/pull/715/commits/18696d1639063636d9a07b905db4f4b0479e8b27) we had several notes displayed inline, separated by with comma + and separators
inline_notes

## After
paragraph-separated-notes

### DIFF
--- a/spec/fixtures/bibliography/book.bib
+++ b/spec/fixtures/bibliography/book.bib
@@ -8,7 +8,5 @@
 	edition = {1st},
 	year = {1962},
 	keywords = {sg012cr8689},
-	annote = { KAW},
-	annote = {CCC MS 470 [p. 9, and referred to (without MS number) throughout]reproductions from CCC MS 470, ff. 125-146, in the appendix},
-	annote = {The following values have no corresponding Zotero field:custom1: 23}
+	annote = { KAW{\textbar} CCC MS 470 [p. 9, and referred to (without MS number) throughout]reproductions from CCC MS 470, ff. 125-146, in the appendix{\textbar} The following values have no corresponding Zotero field:custom1: 23}
 }
--- a/app/controllers/catalog_controller.rb
+++ b/app/controllers/catalog_controller.rb
@@ -180,7 +180,7 @@ class CatalogController < ApplicationController
     config.add_index_field 'ref_type_ssm', label: 'Reference Type'
     # This was added for the Feigbenbaum exhibit.  It includes any general <note> from
     #  the MODs that do not have attributes.  It is used for display and is not facetable.
-    config.add_index_field 'general_notes_ssim', label: 'Notes'
+    config.add_index_field 'general_notes_ssim', label: 'Notes', helper_method: :paragraph_wrap
     config.add_index_field 'collection_with_title', label: 'Collection', helper_method: :document_collection_title
     # "fielded" search configuration. Used by pulldown among other places.
     # For supported keys in hash, see rdoc for Blacklight::SearchFields
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -22,4 +22,9 @@ module ApplicationHelper
       image_tag 'iiif-drag-n-drop.svg', width: '40px', alt: 'IIIF Drag-n-drop'
     end
   end
+
+  def paragraph_wrap(options = {})
+    return if options[:value].blank?
+    safe_join(options[:value].map { |value| content_tag('p', value) })
+  end
 end
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -25,6 +25,12 @@ module ApplicationHelper
 
   def paragraph_wrap(options = {})
     return if options[:value].blank?
-    safe_join(options[:value].map { |value| content_tag('p', value) })
+    safe_join(options[:value].map do |value|
+      if value.start_with?('<p>')
+        value.html_safe
+      else
+        content_tag('p', value.html_safe)
+      end
+    end)
   end
 end
--- a/app/models/bibliography.rb
+++ b/app/models/bibliography.rb
@@ -42,6 +42,7 @@ class Bibliography
     # BibTeX.parse `filter: latex` doesn't handle these cases correctly
     bibtex_data.gsub!(/\\textbackslash/i, '')
     bibtex_data.gsub!(/\\textit/i, '')
+    bibtex_data.gsub!(/\\textbar/i, '|')
 
     bibtex_data
   end
--- a/spec/features/bibliography_resource_integration_spec.rb
+++ b/spec/features/bibliography_resource_integration_spec.rb
@@ -101,6 +101,7 @@ RSpec.describe 'Bibliography resource integration test', type: :feature do
       end
 
       it 'has annotations' do
+        expect(document['general_notes_ssim'].first).to eq 'ELB'
         expect(document['general_notes_ssim'].last).to match(/^The following values/)
       end
     end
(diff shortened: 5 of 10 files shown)


camillevilla:
## Option 1: `<ul>`
### multiple notes
unordered-list-notes

### single note
single-note

## Option 2: `<p>`
paragraph-separated-notes

camillevilla:
Per discussion with @caaster and @blalbrit on 10/11, we'll be using paragraph separators for `general_notes_ssim` to generalize this field for other exhibits (notably, values of /mods/note without a displayLabel #543)

ggeisler:
@camillevilla I like option 2 for presenting multiple notes, but do we have to worry about a given note having embedded paragraphs? If so, it might make it difficult to distinguish two different notes from a note with multiple paragraphs.

If notes with multiple paragraphs is not something we'll see, then I like the `<p>` approach.

ggeisler:
@caaster @blalbrit I saw after my comment above that you had made the decision to use `<p>` for the separator. See my comment about a potential concern (that might not be relevant) about using a `<p>`.

blalbrit:
Saw your comment @ggeisler - I think it's a good concern. We were looking at #543 as a larger "general notes" use-case that needs to be supported. The multi-paragraph note may be a thing for the larger use-case - we would need to do a bit more analysis.

@caaster - do you have more info from the MODS note domain that might cause us to reconsider the <p> approach?

caaster:
Yes, Gary's comment about multiple paragraphs in a single note is an excellent one. Right now I can think of at least one use case where I think that occurs, but there could be more. Let me ask Arcadia to look at this ticket.

arcadiafalcone:
Yes, the multiparagraph note question comes up fairly often. I've been recommending the use of <br> rather than <p>, though I can't answer for legacy data.

If #544 results in additional note index fields, they should behave in the same way.

Also relevant: #697 would apply the paragraph separator to all fields.

camillevilla:
Thanks for the context, @arcadiafalcone! I'll experiment with a few different fixtures (multi-paragraph notes with `<br />` and `<p>` tags) and post samples here.

spec/fixtures/bibliography/book.bib
```diff
@@ -10,5 +10,5 @@ @book{http://zotero.org/groups/1051392/items/TXXUJDG2
 	keywords = {sg012cr8689},
 	annote = { KAW},
 	annote = {CCC MS 470 [p. 9, and referred to (without MS number) throughout]reproductions from CCC MS 470, ff. 125-146, in the appendix},
```
drh-stanford: You're removing the multiple `annote` keys in other fixtures. Do you want to do the same for this fixture, or do you have a spec in mind for handling this case?

lib/traject/bibtex_config.rb
```diff
@@ -38,7 +38,7 @@
 to_field 'issue_ssm', extract_bibtex_field(:issue)
 to_field 'pages_ssm', extract_bibtex_field(:pages)
 to_field 'doi_ssim', extract_bibtex_field(:doi)
-to_field 'general_notes_ssim', extract_bibtex_field(:annote, split: '|', trim: true)
+to_field 'general_notes_ssim', extract_bibtex_field(:annote, split: '\\textbar', trim: true)
```
drh-stanford: If you want, you can make this change in the preprocessing code so do a `gsub('\\textbar', '|')` here -- https://github.com/sul-dlss/exhibits/blob/master/app/models/bibliography.rb#L44

camillevilla: Thanks for the tip! That worked perfectly.

spec/fixtures/bibliography/book.bib
camillevilla: Nope! Just a typo 😅

camillevilla:
I've done a quick fix to handle the possibility of both `<p>` tags and `<br />` tags in the notes. 

### Unescaped HTML with paragraph separators
unescaped_html

### Escaped HTML with paragraph separator
Note the gap caused by double paragraph tags
double-paragraph tag

### Fix
single-wrap

This implementation skips wrapping the content in a `<p>` tag if it there is already a `<p>` tag present in the note.

ggeisler:
@camillevilla I'm not sure if your focus here was solely on dealing with spacing issues when a note has embedded `<p>` or `<br>` tag. If so, great.

But I'm unsure if there is more work planned to focus on the presentation of multiple notes? If there is no further work planned on the presentation, I'm concerned, based on your screenshot above, that this is not a complete solution. Let me know and I'll provide more details of my concern if necessary.

camillevilla:
@ggeisler this work was just focused on the spacing issues. Looking back at your comment, I see that I missed your finer point about distinguishing two different notes from a note with multiple paragraphs 😥

I'll make another ticket to make sure multiple notes get handled with a `<ul>` tag.

ggeisler:
@camillevilla Okay, great. I know a `ul` in the metadata list is not particularly elegant; we could experiment with styling to make it a bit more subtle, such as making the bullets a gray rather than the darker default color, which would deemphasize them while retaining their value as an item indicator.

PR #716, opened 2017-10-10
permit up to 1000 hits for JSON API

### LINKED ISSUE #700: Is the bibliography display on show page limited to a certain number of objects?
Per conversation with @mejackreed below, it looks like we're capped at 100 bibliographic items on a show page. We'll need to increase this number to accommodate some of the more studied manuscripts. CCC 286 has ~223 citations, for instance. I'm not sure which object has the largest bibliography - but even the largest will grow over time, so we might want a good buffer.

Original question convo below:

Not quite sure how to phrase this - but when I look at some of the bibliography lists, they appear truncated.

For instance: https://exhibits.stanford.edu/parker-2/catalog/mk707wk3350

The bibliography list ends with M.R. James, in this example, when it should go all the way to Youngman (if we're looking at the Parker 1.x version - https://parker-stanford-edu.stanford.idm.oclc.org/parker/actions/bibliography_search_results.do?journal=&primarySourceRef=286&author=&title=&sort=&shortTitle=&comment=&date=&referenceTypes=)

 Acceptance criteria:
 - [ ] Full display of all bibliography items for a given manuscript (not capped)

drh-stanford:
This PR fixes #700. It overrides `SearchBuilder#rows` to permit higher max_per_page values for the JSON API.  

Note that I had problems getting the SearchBuilder specs to work with the advanced search configuration so I have a solution for it but it's not ideal.

Note to try this out, load the entire Parker bibliography and then visit http://localhost:3000/test/catalog/mk707wk3350. The full bibliography for that manuscript (200+ references) should show up.

### After
long-bibliography mov

### DIFF
--- a/app/models/search_builder.rb
+++ b/app/models/search_builder.rb
@@ -6,4 +6,19 @@ class SearchBuilder < Blacklight::SearchBuilder
   include BlacklightHeatmaps::SolrFacetHeatmapBehavior
 
   include Spotlight::AccessControlsEnforcementSearchBuilder
+
+  ##
+  # modify JSON API behavior to limit the `rows` (or `per_page`) parameter
+  # to `max_per_page_for_api` (default: 1,000). if `rows` is not provided
+  # we use `default_per_page`
+  def rows(value = nil)
+    return super if value || blacklight_params[:format] != 'json'
+
+    @rows = [:rows, :per_page].map { |k| blacklight_params[k] }.reject(&:blank?).first
+    @rows = if @rows.blank?
+              blacklight_config.default_per_page
+            else
+              [@rows.to_i, (blacklight_config.max_per_page_for_api || 1_000)].min # ensure under max
+            end
+  end
 end
--- a/spec/models/search_builder_spec.rb
+++ b/spec/models/search_builder_spec.rb
@@ -1,0 +1,81 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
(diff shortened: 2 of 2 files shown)


app/models/search_builder.rb
```diff
@@ -2,8 +2,17 @@
...
+                                     add_advanced_search_to_solr
+                                     json_api_max_per_page)
   include BlacklightHeatmaps::SolrFacetHeatmapBehavior
 
   include Spotlight::AccessControlsEnforcementSearchBuilder
+
+  JSON_API_MAX = 1_000
+  def json_api_max_per_page(solr_parameters)
+    return unless blacklight_params[:format] == 'json'
+    solr_parameters[:rows] = (blacklight_params[:rows] || blacklight_config.max_per_page).to_i
```
cbeer: Should this be `default_per_page` instead of `max_per_page`?

Maybe, instead of adding another processor, you'd like to override `#rows` instead? Here's the implementation of `rows` upstream (which is used to feed the `rows` parameter in the solr query):

https://github.com/projectblacklight/blacklight/blob/v6.11.2/lib/blacklight/search_builder.rb#L201-L203

drh-stanford: Ok, I'll give that a try.

drh-stanford:
Ok I've rewritten the implementation to override the `rows` method in our SearchBuilder.

PR #717, opened 2017-10-11
do not save search in history if JSON API

### LINKED ISSUE #701: Figure out a way not to store a search result to reduce pollution of search session
Acceptance criteria (this shouldn't happen):
 - After clicking on a show page that has a bibliography clicking "Search results" returns a user to an incorrect search result.

drh-stanford:
This PR fixes #701. It overrides the `start_new_search_session?` method for any JSON API search request and sets it to `false` so that the `find_search_session` in Blacklight returns `nil` and thus the `Search` object never gets created.

Works for any JSON API request, so the heatmaps view also works.

### After 

The "Back to Search" and "Search Results" actions have the expected behavior. The bibliography on the manuscript show page uses the JSON API asynchonously.

search_history mov

### DIFF
--- a/app/controllers/catalog_controller.rb
+++ b/app/controllers/catalog_controller.rb
@@ -267,4 +267,9 @@ class CatalogController < ApplicationController
     config.add_sort_field 'author_sort asc, title_sort asc', label: 'author'
     config.add_sort_field 'title_sort asc, pub_year_isi desc', label: 'title'
   end
+
+  # JSON API queries should not trigger new search histories
+  def start_new_search_session?
+    super && params[:format] != 'json'
+  end
 end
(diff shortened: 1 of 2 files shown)


cbeer:
Is it possible to override `start_new_search_session?` here?
  https://github.com/projectblacklight/blacklight/blob/afd4741a130f0ff75164884077246317b1e8e0cc/app/controllers/concerns/blacklight/search_context.rb#L64

drh-stanford:
thanks i'll give that a try...

drh-stanford:
So that worked and I've updated the PR to reflect that we're overriding the `start_new_search_session?` method to handle this issue. Thanks!

PR #735, opened 2017-10-12
Remove media-query to help results thumbnail alignment.

jkeck:
Nice catch @camillevilla!  Does this look better?

## Before
before

## After
after

### DIFF
--- a/app/assets/stylesheets/modules/blacklight_overrides.scss
+++ b/app/assets/stylesheets/modules/blacklight_overrides.scss
@@ -81,10 +81,6 @@
       height: 0;
       margin: 0;
       padding: 0;
-
-      @media (max-width: $screen-sm-min) {
-        height: auto;
-      }
     }
   }
 


jkeck:
I'm having a bit of trouble reproducing the title running into the thumbnail issue though, so perhaps you can help me figure that one out.

camillevilla:
@jkeck I'm not sure why, but I'm having difficulty reproducing the behavior from your screenshots.
I'm getting a lot more of this:
### Small < x <  Medium
screen shot 2017-10-12 at 4 09 53 pm

Some weird stuff is happening with the [offsets](https://github.com/sul-dlss/exhibits/blob/2020e85609450dc827e68f6bbc07adbf898619a6/app/assets/stylesheets/modules/blacklight_overrides.scss#L72) in `.document-metadata`. Would it be possible to move these out of `@extend`s for easier debugging?

PR #740, opened 2017-10-13
add Manuscript Number to indexing

### LINKED ISSUE #721: Index and make searchable manuscript number
see `<shelfLocator>` in https://purl.stanford.edu/cf386wt1778.mods for where this is currently.

- [x] Index metadata
- [x] Design fielded search UI
- [x] Support fielded search
- [x] Support "exact match" searches for manuscript number #778

drh-stanford:
This PR is connected to #721 and it only adds the manuscript_number_ssim Solr field.

### DIFF
--- a/spec/models/spotlight/dor/indexer_spec.rb
+++ b/spec/models/spotlight/dor/indexer_spec.rb
@@ -739,4 +739,36 @@ describe Spotlight::Dor::Indexer do
       end
     end # add_object_full_text
   end # full text indexing concern
+
+  context 'Parker specific indexing' do
+    let(:manuscript_number_field) { 'manuscript_number_ssim' }
+
+    describe '#add_manuscript_number' do
+      before do
+        subject.send(:add_manuscript_number, resource, solr_doc)
+      end
+
+      it 'without a manuscript number, expect blank' do
+        expect(solr_doc[manuscript_number_field]).to be_blank
+      end
+
+      context 'with a manuscript number' do
+        # e.g. from https://purl.stanford.edu/cf386wt1778.mods
+        let(:modsbody) do
+          <<-EOF
+            <location>
+              <physicalLocation type="repository">
+                UK, Cambridge, Corpus Christi College, Parker Library
+              </physicalLocation>
+              <shelfLocator>MS 69</shelfLocator>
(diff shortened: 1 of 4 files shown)


jkeck:
Of of the questions I had about the acceptance criteria of these tickets is what does "and make searchable" mean?  If it just means that it's part of the field list that is searched under general keyword search, then that's fine.  If it means that it is available as an option in the search drop down, then we may either need to add that to this PR or split that out into a diff. ticket.

drh-stanford:
Yeah, I think this PR is just going to be the indexing. The searchability may need design work, or at least a strategy on whether these Parker fields are available across the whole site. In other words, if I add a config search field then it'll be available in all exhibits:

```

**Comment by drh-stanford** [comment]
I changed this PR to not close the main ticket. I think we need to discuss more _how_ these fields are searchable and whether it's specific to the Parker exhibit. I'll comment on that in the main ticket. If I just add a config search field, it'll show up in all the exhibits as I understand it:

```
    config.add_search_field('manuscript_number') do |field|
      field.label = 'Manuscript Number'
      field.include_in_advanced_search = false
      field.solr_parameters = { defType: 'lucene' }
      field.solr_local_parameters = {
        df: 'manuscript_number_ssim'
      }
    end
```

File: spec/models/spotlight/dor/indexer_spec.rb [hunk 1]
```diff
@@ -1,5 +1,7 @@
 require 'rails_helper'
 
+# rubocop: disable Metrics/BlockLength
```
jkeck: What would you think about disabling this for the `spec` directory entirely?

camillevilla: +1 that notion ☝️

drh-stanford: Fine with me. The rubocop todo has a huge number for this cop already.

drh-stanford:
Ok, I've made the changes to Rubocop's configuration. The PR is ready for review.

camillevilla:
@drh-stanford this indexing work looks good to me. We can split off another implementation ticket from #721 once the design work in #743 has been reviewed.

PR #744, opened 2017-10-13
refine multiple notes styling

### LINKED ISSUE #742: Refine display of multiple notes
Design: #605 
Discussion / Context: PR #715 

Let's refine how multiple values in the `general_notes_ssim` field are displayed. Values are currently separated by `<p>` tags via a Blacklight helper. We need a good way to distinguish two different notes from a note with multiple paragraphs. 

Example entry for `general_notes_ssim`:
```
['Here's a single paragraph note', 
'<p>Here's a note with embedded paragraph tags</p><p>Here's the second paragraph</p>']
```

Notes from @ggeisler:
>we could experiment with styling to make it a bit more subtle, such as making the bullets a gray rather than the darker default color, which would deemphasize them while retaining their value as an item indicator.

camillevilla:
Closes #742 

- [x] replace `<p>` separators with a `<ul>` 
- [x] differentiating styling suggestions from @ggeisler 

## Before
before_full_note

## After
first_iteration_full_note #

### DIFF
--- a/spec/helpers/application_helper_spec.rb
+++ b/spec/helpers/application_helper_spec.rb
@@ -17,9 +17,11 @@ describe ApplicationHelper, type: :helper do
     end
   end
 
-  describe '#paragraph_wrap' do
-    it 'separates multivalued fields into paragraphs' do
-      expect(helper.paragraph_wrap(value: %w(a b))).to eq '<p>a</p><p>b</p>'
+  describe '#notes_wrap' do
+    let(:output) { '<ul class="general-notes"><li>a</li><li><p>b</p><p>c</p></li><li>d</li></ul>' }
+
+    it 'permits embedded HTML and handles multivalued notes as an unordered list' do
+      expect(helper.notes_wrap(value: %w(a <p>b</p><p>c</p> d))).to eq output
     end
   end
 end
--- a/app/assets/stylesheets/modules/bibliography.scss
+++ b/app/assets/stylesheets/modules/bibliography.scss
@@ -27,3 +27,7 @@
 .btn-view-on-zotero {
   margin-right: 10px;
 }
+
+.general-notes {
+  padding-left: 15px;
+}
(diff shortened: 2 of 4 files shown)


camillevilla:
@ggeisler It looks like changing bullet colors is a more involved process than I thought it would be. It requires [some hacky CSS](https://www.w3.org/Style/Examples/007/color-bullets.en.html). 

`::before` properties for `<li>` don't play well with embedded `<p>` tags. 
screen shot 2017-10-13 at 2 52 41 pm

Getting notes to play well will require something like 
(a) stripping out `<p>` tags (or at least just the first occurrence per note)
(b) stripping out all HTML tags and ONLY whitelisting `<br />` tags (which might not play well with legacy metadata)

An example of option (a)
screen shot 2017-10-13 at 2 42 18 pm

ggeisler:
@camillevilla Yeah, those options seem way too involved for what we're trying to get. I'm not even sure we need to do any additional styling for the note items. That was just an off-the-cuff idea to mitigate the somewhat less than ideal scenario of presenting a `ul` within a metadata value element. But I'm not sure we really need to do anything beyond what we get by default. What do you think of the default output?

camillevilla:
@ggeisler the default output looks good to me! And this implementation is a little cleaner than my original attempt.

ggeisler:
@camillevilla Okay, cool. Thanks for experimenting with the styling anyway!

spec/helpers/application_helper_spec.rb
```diff
@@ -17,9 +17,9 @@
     end
   end
 
-  describe '#paragraph_wrap' do
+  describe '#notes_wrap' do
     it 'separates multivalued fields into paragraphs' do
-      expect(helper.paragraph_wrap(value: %w(a b))).to eq '<p>a</p><p>b</p>'
+      expect(helper.notes_wrap(value: %w(a b))).to eq '<ul class="general-notes"><li>a</li><li>b</li></ul>'
```
drh-stanford: do you want to add a test here for multi-paragraph notes? my understanding is that the `li` implementation is to accomodate them.

camillevilla: ah! good catch. I've updated the test description and added some embedded paragraph tags. Just waiting to appease the Rubocop now...

Issue #723, opened 2017-10-12
Index and make searchable text title


mejackreed:
See entry in https://purl.stanford.edu/cf386wt1778.mods `<tableOfContents>`

- [x] Index
- [x] Display
- [x] Searchable

drh-stanford:
@blalbrit what should the label be for text titles? `Table of contents`?

blalbrit:
@drh-stanford - yes, I think `Table of contents` - but again, would be helpful to have @ggeisler 's thoughts.

ggeisler:
@blalbrit @drh-stanford Sure, given the element is `<tableOfContents>`, it also makes sense to me that the label be `Table of Contents`.

(Though -- and this might not be relevant to this ticket -- I don't completely understand how this element is used. In the example document given above, there is only one `<tableOfContents>` element. Is there only ever one such element per document? If there can be more than one, are they nested somehow?

It just seems a little unusual to refer to a single text string as a "table of contents" since in normal usage a table of contents is a list of entries, not a singular entry. But I could very well just not understand the relevant context here.)

blalbrit:
@ggeisler @drh-stanford For modeling purposes, I used our simplest manuscript to build out the first table of contents. There can be hundreds of entries there depending on the manuscript. I think there are some questions about display of this element across the board in Stanford MODS display...

arcadiafalcone:
tableOfContents is not a repeatable field in MODS, and does not have any child elements. This may be a case of doing internal `<br/>` tags for displaying each entry on its own line.

ggeisler:
Okay, great, then it sounds like going with `Table of Contents` for the label is the right answer.

camillevilla:
Indexing and basic display was added to the show page by #746. It will be separated by the usual commas and `and`. We might need some longer examples of `<tableOfContents>` down the line.

Test druid: cf386wt1778
![screen shot 2017-10-18 at 4 29 50 pm](https://user-images.githubusercontent.com/5402927/31747426-9f0c3344-b421-11e7-835d-017180806d85.png)

PR #746, opened 2017-10-17
add tableOfContents to MODS indexing

drh-stanford:
This PR is connected to #723. It indexes the MODS `tableOfContents` into `text_titles_ssim`.

### DIFF
--- a/app/models/spotlight/dor/indexer.rb
+++ b/app/models/spotlight/dor/indexer.rb
@@ -281,12 +281,19 @@ module Spotlight::Dor
     concerning :ParkerIndexing do
       included do
         before_index :add_manuscript_number
+        before_index :add_text_titles
       end
 
       def add_manuscript_number(sdb, solr_doc)
         manuscript_number = sdb.smods_rec.location.shelfLocator.try(:text)
         return if manuscript_number.blank?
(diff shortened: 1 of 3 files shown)


app/models/spotlight/dor/indexer.rb
```diff
@@ -278,6 +278,18 @@ def object_level_full_text_filenames(sdb)
       end
     end
 
+    concerning :ParkerIndexing do
+      included do
+        before_index :add_text_titles
+      end
+
+      def add_text_titles(sdb, solr_doc)
+        text_titles = sdb.smods_rec.tableOfContents.try(:content)
```
drh-stanford: I'm not sure whether we need to further parse the tableOfContents into multiple lines (i.e., `split(/\n/)`).

camillevilla:
👍 works for me
Test druid: cf386wt1778
screen shot 2017-10-18 at 4 29 50 pm

app/models/spotlight/dor/indexer.rb
camillevilla: Just read the discussion in #723 and I'm not sure either. I'll ask for more examples over there.

PR #766, opened 2017-10-18
Bugfix: only add IIIF logos to SDR objects with manifests

### LINKED ISSUE #745: The IIIF logo is appearing on objects that are not images
Currently, the IIIF logo appears on all objects, regardless of content type.

The IIIF logo should only appear on the following content types:
- image
- book
- map

camillevilla:
Fixes #745 

This PR:
- fixes a bug that was adding IIIF drag-n-drop logos to invalid SDR objects (e.g. a dataset, an email archive)
- adds a check to our SDR indexing pipeline: is the generated IIIF manifest link valid? (i.e. does the URL return a 200 status code)
- removes Rubocop BlockLength limits from our `dor/indexer_spec.rb`

### no manifest
py883nd2578
rj585xk6315

### existing manifest
sg012cr8689

note: the screenshots below are made up of three separate, cropped images 
## Before
data_before
email_before
valid iiif object before

## After
dataset_after
email_after
valid iiif object

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -14,6 +14,10 @@ AllCops:
 Rails:
   Enabled: true
 
+Metrics/BlockLength:
+  Exclude:
+  - 'spec/models/spotlight/dor/indexer_spec.rb'
+
 Metrics/LineLength:
   Max: 120
   Exclude:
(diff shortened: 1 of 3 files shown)


camillevilla:
thanks for the tip, @drh-stanford. I've updated the spec and stubbed out Faraday responses for both valid and invalid URLs. I ended up using `instance_double` to appease RuboCop.

PR #770, opened 2017-10-19
Add missing migration from Spotlight that migrates sidecar taggings…

### LINKED ISSUE #753: Reindexing removes custom tagged items from browse categories
This afternoon I reindexed the Travel Through Time exhibit: https://exhibits.stanford.edu/time-travel-japan. AFTER reindexing (and I didn't check before) -- I noticed that several browse categories were displaying with `0` items: https://exhibits.stanford.edu/time-travel-japan/browse. This exhibit used custom tags to help create several browse categories. Total items for each custom tag are still displaying on Dashboard > Curation > Tags : https://exhibits.stanford.edu/time-travel-japan/tags -- yet when I execute a search on, for example: `Temple Precincts` (a browse category, and a custom tag with 46 items), the search result comes up empty.

I was concerned that perhaps my reindexing of the exhibit had caused this to occur (it should not have, but I was worried about any recent work that could have caused this indexing issue to occur). So, I went to -stage (where unfortunately, we do not have an instance of the Travel Through Time exhibit) -- but I found an exhibit there that has lots of custom tags used to create many browse categories (https://exhibits-stage.stanford.edu/harrison). I took before screen shots of the browse categories (see attachments), which now display many `0`results after reindexing. I also took before-reindexing screen shots of all the custom tags. 

Expected behavior: Custom tags & their relationships should be retained when reindexing an exhibit.

screen shot 2017-10-17 at 4 34 26 pm
screen shot 2017-10-17 at 4 34 38 pm
screen shot 2017-10-17 at 4 33 17 pm
screen shot 2017-10-17 at 4 33 05 pm
screen shot 2017-10-17 at 4 33 25 pm

jkeck:
… from SolrDocument id to Sidecar (AR) id.

Fixes #753 

I haven't tested this fix out yet, so somebody should deploy this to stage and re-index the harrison exhibit to see if the traggings remain after re-index.

### DIFF
--- a/db/migrate/20171017223044_migrate_tags_to_sidecars.spotlight.rb
+++ b/db/migrate/20171017223044_migrate_tags_to_sidecars.spotlight.rb
@@ -1,0 +1,18 @@
+# This migration comes from spotlight (originally 20170803152134)
+class MigrateTagsToSidecars < ActiveRecord::Migration[5.0]
+  def up
+    Spotlight::SolrDocumentSidecar.reset_column_information
+    ActsAsTaggableOn::Tagging.reset_column_information
+
+    ActsAsTaggableOn::Tagging.where(taggable_type: 'SolrDocument', tagger_type: 'Spotlight::Exhibit').find_each do |e|
+      sidecar = Spotlight::SolrDocumentSidecar.find_or_create_by(document_id: e.taggable_id, document_type: 'SolrDocument', exhibit_id: e.tagger_id)
+      e.update(taggable: sidecar)
+    end
+  end
+  
+  def down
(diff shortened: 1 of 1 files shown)


jkeck:
Also, not sure how we can codify this or make it more explicit, but anytime we update Spotlight we need to run `rake spotlight:install:migrations`.

This is the 2nd time we've been burnt because we didn't do that.

drh-stanford:
I got the following message when trying to run that rake task:

```
$ RAILS_ENV=production bundle exec rake spotlight:install:migrations
WARNING: Rubocop was not found and could not be required.
NOTE: Migration 20170821165811_change_index_status_to_longblob.rb from spotlight has been skipped. Migration with the same name already exists.
```

If we need to run that rake task we can add it to our Capistrano deploy script using `test` rather than `exec`.

PR #794, opened 2017-10-24
enforce IIIF manifest availability by using contentMetadata type

### LINKED ISSUE #745: The IIIF logo is appearing on objects that are not images

drh-stanford:
This PR fixes #745 by reducing the time indexing takes (avoiding a fetch on the IIIF manifest to determine validity).

It uses the contentMetadata type attribute to determine whether a resource will have a valid IIIF manifest. The `iiif_manifest_url_ssi` is always present as it's just the location of where the manifest _should_ be. The affected specs are both in the SolrDocument and the views.

cc: @camillevilla @mejackreed

### DIFF
--- a/app/models/concerns/manifest_concern.rb
+++ b/app/models/concerns/manifest_concern.rb
@@ -3,7 +3,9 @@
 # convenient accessors for IIIF manifests embeded in a SolrDocument
 module ManifestConcern
   def manifest
-    fetch('iiif_manifest_url_ssi', nil)
+    manifest_url = fetch('iiif_manifest_url_ssi', nil)
+    return if manifest_url.blank? || !manifest_available?
+    manifest_url
   end
 
   def exhibit_specific_manifest(custom_manifest_pattern)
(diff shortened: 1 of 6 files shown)


app/models/concerns/manifest_concern.rb
```diff
@@ -2,8 +2,15 @@
...
 module ManifestConcern
+  VALID_IIIF_CONTENT_TYPES = %w(image manuscript map book).freeze
+
+  # even if a URL is present we need to check the contentMetadata type
+  # to determine whether the manifest will resolve at runtime.
   def manifest
-    fetch('iiif_manifest_url_ssi', nil)
+    manifest_url = fetch('iiif_manifest_url_ssi', nil)
+    return if manifest_url.blank?
+    return unless VALID_IIIF_CONTENT_TYPES.include?(first('content_metadata_type_ssm'))
```
mejackreed: I wonder if this line could be extracted to another method if that might be useful elsewhere?

drh-stanford: Ok, I did an extraction for it to a method called `manifest_available?`

spec/views/catalog/_exhibits_document_header_default.html.erb_spec.rb
```diff
@@ -17,7 +17,10 @@
 
   describe 'IIIF Drag-n-Drop Icon' do
     context 'when the document has a manifest' do
-      let(:document) { SolrDocument.new(iiif_manifest_url_ssi: 'https://purl.stanford.edu/bc853rd3116/iiif/manifest') }
+      let(:document) do
+        SolrDocument.new(iiif_manifest_url_ssi: 'https://purl.stanford.edu/bc853rd3116/iiif/manifest',
+                         content_metadata_type_ssm: %w(image))
+      end
```
camillevilla: @drh-stanford could you add a case for when the document has a manifest but does not have a valid `content_metadata_type_ssm`?

drh-stanford: Ok, I added that test.

PR #795, opened 2017-10-24
[WIP] Add a separate field for each value of /mods/note/@displayLabel in an exhibit

camillevilla:
Addresses part of #544 

Sample record:
[https://purl.stanford.edu/cf386wt1778.mods](https://purl.stanford.edu/cf386wt1778.mods)

## [WIP]
- [x] write some tests 
### ~~Option 1 - index notes as separate fields~~
- [x] Get SOME working solr_fields indexed
- [x] add both `<relatedItem>` and `<physicalDescription>` notes
- [ ] use displayLabel for the field label
- [x] figure out how to handle multiple notes of the same type 
e.g. 
```
<note displayLabel="Height" type="dimensions">300 mm</note>
<note displayLabel="Width" type="dimensions">215 mm</note>
<note displayLabel="Dimensions" type="dimensions">12 x 8.7 in</note>
```
### Option 2 - index notes in the same field, but display separately
- [x] Throw notes into a `labeled_notes_ssim` field and add a helper for displaying things as separate fields
- [x] Hide `Labeled notes`
- [ ] Implement #799

### DIFF
--- a/app/assets/stylesheets/modules/labeled_notes.scss
+++ b/app/assets/stylesheets/modules/labeled_notes.scss
@@ -1,0 +1,6 @@
+dd,
+dt {
+  &.blacklight-labeled_notes_ssim {
+    display: none;
+  }
+}
--- a/app/assets/stylesheets/sul_theme.scss
+++ b/app/assets/stylesheets/sul_theme.scss
@@ -7,3 +7,4 @@
 @import "modules/blacklight_overrides";
 @import "modules/spotlight_overrides";
 @import "modules/blacklight_heatmaps_overrides";
+@import "modules/labeled_notes";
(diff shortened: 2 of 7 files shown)


camillevilla:
Here's a stab at Option 2. Everything from `Rubric 1` through `Foliation` is actually a child of `Labeled Notes` / `labeled_notes_ssim`. Note that these fields are not facetable (@jkeck suggested a whitelist approach in #544, but I haven't implemented that here). 

@blalbrit do you have any feedback on this display? 

## Before
screen shot 2017-10-25 at 3 31 17 pm

## After
annotated

blalbrit:
@camillevilla I like that these are displaying a much richer set of metadata. I would suggest suppressing the "Labeled Notes" label, if possible, as that won't mean much to our end users. If I understand your note correctly, you are offering an on/off toggle for all labeled notes. If so, Parker only requires "on" - we don't want to toggle off I don't think, unless @ggeisler has other thoughts.

For Parker, I'm comfortable with not including faceting in the requirement. This approach satisfies our user needs.

blalbrit:
To build on my previous comment - after talking with stakeholders, I reverse my position on toggling on and off all notes, @camillevilla - would be happy to talk through that choice if you want.

camillevilla:
I was able to hide `Labeled Notes` with CSS. 
cambridge__corpus_christi_college__ms_69__gregory_the_great__homiliae_xl_in_euangelia__-_test_-_stanford_university_libraries

But it does look like the show page is starting to drag down...
show-page

PR #806, opened 2017-10-31
Index related IIIF annotations for SDR objects.

### LINKED ISSUE #757: Enable feature flag for annotation indexing.
Depends on #756 (ReindexJob)

Acceptance criteria:

- [ ] Annotation indexing should only be enabled for Parker exhibit

Story #714

### LINKED ISSUE #805: Hook up IIIF Annotation List indexing to DorHarvester
At some point in DorHarvester (maybe on create? but we need all of the known druids) we should create an independent IIIF Dor harvester. Split out of #758 

This thing should have the responsibility of:
 - Be able to get a set of druids to do stuff with
 - For all druids, understand a manifest url(and handle custom manifests)
 - Enqueue a indexer for each druid to grab the manifest and index its content.

mejackreed:
Closes #805 
Closes #757 

To see this in action, set a custom mani url for an exhibit to : `https://dms-data.stanford.edu/data/manifests/Parker/{id}/manifest.json`

Then index one (or all) of these:

- `fh878gz0315`
- `bg021sq9590`
- `ft757ht3699`

## Notes on performance

In my testing I was able to see this successfully work using the default `:async` ActiveJob adapter. However, it happens much slower this way and usually takes a bit. For faster performance use the `:sidekiq` adapter.

I also tested this with the entire Parker collection in development by indexing the collection druid on 
a fresh exhibit. Things seemed to work ok, but many failed jobs due to expected SQLite timeouts. Everything ended up completing within an hour.

TODO's

 - [x] feature flag this / exhibit
 - ~~[ ] reindex count not being updated~~ Split out to #808


app/models/iiif_canvas_indexer_enqueuer.rb
```diff
@@ -0,0 +1,41 @@
+# frozen_string_literal: true
+
+##
+# For a given object in an exhibit, gets the correct manifest, and enqueues
+# canvas/page indexing jobs from that manifest.
+class IiifCanvasIndexerEnqueuer
```
cmh2166: I feel like this class is where the overlap with how the Traject+ Manifest > Canvas > AnnoList > Anno capture, mapping, and indexing ('transform') could be reused. I don't think how this PR does the work should change for that reason, but it would be a nice way to take some of this IIIF resource class logic and abstract out in the same Traject+ code for other semantic logic (like MODS, Bibtex, other). Also could help solve some of the iteration issues, if we can call Traject at a Anno-level (for example) within this, but again, tomato / tomahtoe.

cmh2166: I've added a IIIF support at various levels issue to the Traject+ repo, fwiw.

app/models/dor_harvester.rb
```diff
@@ -51,6 +51,11 @@ def indexable_resources
 
   def on_success(resource)
     RecordIndexStatusJob.perform_later(self, resource.bare_druid, ok: true)
+    IndexRelatedContentJob.perform_later(self, resource.bare_druid) if index_related_content?
+  end
+
+  def index_related_content?
+    FeatureFlags.new.for(exhibit.slug).index_related_content?
```
jkeck: You can drop the `new` here if you want: https://github.com/sul-dlss/exhibits/blob/502f4496ffc9c0c2a981912ed71f44763692acef/app/models/feature_flags.rb#L28-L29

app/models/iiif_canvas_indexer_enqueuer.rb
```diff
@@ -0,0 +1,41 @@
...
+  def initialize(exhibit, druid)
+    @exhibit = exhibit
+    @druid = druid
+  end
+
+  def enqueue_jobs
+    canvases.each do |canvas|
+      canvas.other_content.each do |other_content|
+        # Don't bother to index unless there is an annotationList content to do so.
+        next unless other_content['@type'] == 'sc:AnnotationList'
```
jkeck: What would you think about making this string a constant at the top of the class?  I have gotten in the habit of trying to keep values like this out of the body of methods...not sure I've ultimately seen the benefit of doing it, but it has seemed cleaner.  Totally up to you!

app/models/iiif_manifest_harvester.rb
```diff
@@ -0,0 +1,22 @@
+require 'iiif/presentation'
```
jkeck: Not sure if we need to worry about `frozen_string_literal` here or not, but I noticed it was added in other new files in this PR.

config/settings/test.yml
```diff
@@ -12,3 +12,4 @@ feature_flags:
   bibliography_resource: true
   test-flag-exhibit-slug:
     test_thing: true
+    index_related_content: true
```
jkeck: Can this just blindly be set to `true` in the test suite?

app/models/iiif_canvas_indexer_enqueuer.rb
mejackreed: 👍

app/models/dor_harvester.rb
mejackreed: 👍

config/settings/test.yml
mejackreed: I think so. The idea here was to only enable for this one exhibit to test the check on it. Should Am I missing another case here?

jkeck: Nah, I think I get what you're saying.  Because it'll ultimately be turned on specifically for parker (initially?) we want to test that the exhibit-specific configuration is what is working?  I didn't catch that this was tested in a way that validated that the feature would only be present in the exhibit it is specifically configured to.

Was just more of a passing question than anything, so 👍

app/models/iiif_manifest_harvester.rb
mejackreed: 👍

app/models/iiif_canvas_indexer_enqueuer.rb
jkeck: Did you want to use the constant in the body of this method?

app/jobs/index_canvas_job.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+##
+# Used to kick off the CanvasResource
+class IndexCanvasJob < ActiveJob::Base
```
cbeer: I'm not sure I'm understanding why this job was extracted from `IiifCanvasIndexerEnqueuer`?

app/models/iiif_canvas_indexer_enqueuer.rb
```diff
@@ -0,0 +1,43 @@
...
+    @exhibit = exhibit
+    @druid = druid
+  end
+
+  def enqueue_jobs
+    canvases.each do |canvas|
+      canvas.other_content.each do |other_content|
+        # Don't bother to index unless there is an annotationList content to do so.
+        next unless other_content['@type'] == ANNOTATION_LIST
+        IndexCanvasJob.perform_later(other_content['@id'], canvas.to_json, exhibit)
```
aeschylus: Recording something that came up in a group review of this: there may not necessarily be an @id  or @type. Manifests may only list `otherContent` as a plain string (no object or @id). Not a problem for us now, but we may want to ask about it.

PR #820, opened 2017-11-02
Adjust responsive behavior in small List view

### LINKED ISSUE #736: Fix Small to Medium responsive behavior for document metadata in List view
Document metadata is too far offset in Small to Medium views (768px to 992px)

### Small < x <  Medium
screen shot 2017-10-12 at 4 09 53 pm

Some weird stuff is happening with the [offsets](https://github.com/sul-dlss/exhibits/blob/2020e85609450dc827e68f6bbc07adbf898619a6/app/assets/stylesheets/modules/blacklight_overrides.scss#L72) in `.document-metadata` after #732

camillevilla:
Closes #736 

This PR adjusts the document metadata offset in Small view (between 768px and 992 px). 

## Before
small_view_before

## After
small_view_after

## Demo
responsive-behavior-after

### DIFF
--- a/app/assets/stylesheets/modules/blacklight_overrides.scss
+++ b/app/assets/stylesheets/modules/blacklight_overrides.scss
@@ -73,7 +73,7 @@
       @extend .col-xs-8;
       @extend .col-md-9;
       @extend .col-xs-offset-3;
-      @extend .col-md-offset-2;
+      @extend .col-sm-offset-2;
     }
 
     .document-thumbnail {


app/assets/stylesheets/modules/blacklight_overrides.scss
```diff
@@ -73,6 +73,7 @@
       @extend .col-xs-8;
       @extend .col-md-9;
       @extend .col-xs-offset-3;
+      @extend .col-sm-offset-2;
       @extend .col-md-offset-2;
```
ggeisler: @camillevilla Do you still need `@extend .col-md-offset-2;` after adding the `offset-2` for `sm`? Generally if you specify something at a smaller breakpoint and don't have a different specification at a larger breakpoint the original specification will also apply to the larger breakpoint. But I haven't gone back to look at what exactly is going on in the applicable view, etc. so perhaps that doesn't apply here.

Ignore if this isn't applicable here. 👍 to the end result in any case.

camillevilla: ah! Good catch, Gary. I was so focused on fiddling with `xs` and `sm` that I forgot about the `md`. I've removed `@extend .col-md-offset-2` and updated the PR.

PR #839, opened 2017-11-07
Add a default thumbnail for annotation records.

### LINKED ISSUE #837: Add default thumbnail for canvas document
default-square-thumbnail-annotation
Split off from #830

jkeck:
Closes #837 

annotation-thumbs


app/helpers/catalog_helper.rb
```diff
@@ -12,7 +12,7 @@ module CatalogHelper
   # @return [Boolean]
   # rubocop:disable Style/PredicateName
   def has_thumbnail?(document)
-    return super unless document.reference?
+    return super unless document.reference? || document.canvas?
```
jkeck: I feel like there is a pattern emerging here but I don't know exactly what to call it yet...

PR #847, opened 2017-11-07
Index DOR objects using traject

cbeer:


lib/traject/dor_config.rb
```diff
@@ -0,0 +1,394 @@
...
+  [
+    # feigenbaum style - full text in .txt named for druid
+    "//contentMetadata/resource/file[@id=\"#{sdb.bare_druid}.txt\"]"
+  ]
+end
+
+# rubocop:disable Metrics/AbcSize, Metrics/CyclomaticComplexity, Metrics/PerceivedComplexity
+def file_ids(resource)
+  ids = []
+  if resource.content_metadata
```
jkeck: Can we avoid this guard clause and bail out earlier?

lib/traject/dor_config.rb
```diff
@@ -0,0 +1,394 @@
...
+    if display_type(dor_content_type(resource)) == 'image'
```
jkeck: What would you think about assigning this to a temp variable, it looks like it's called again in the `elseif` below.

lib/traject/adjust_cardinality.rb
```diff
@@ -0,0 +1,47 @@
...
+
+    attr_reader :source
+
+    def call
+      flatten_top_level(source)
+    end
+
+    def flatten_top_level(attributes)
+      flatten = %w(id modsxml access_facet building_facet druid url_fulltext display_type collection_type)
+      flatten += %w(author_1xx_search)
```
jkeck: Should we be using `<<` / `concat` to prevent us from unnecessarily creating a new object each time?

lib/traject/dor_config.rb
cbeer: This method was pulled almost directly from gdor-indexer; I suspect in the next pass we'll throw the whole field away.

lib/traject/adjust_cardinality.rb
cbeer: I think a next step for this code is to stop adjusting cardinality; Solr doesn't care, so this is only to create perfect parity with the current indexing.

lib/traject/dor_config.rb
jkeck: Ah, okay, I was thinking it was invented for this purpose (given that I didn't see it being removed in this PR), but that makes sense if it was from `gdor-indexer`.

lib/traject/adjust_cardinality.rb
jkeck: I'm not sure I understand what that has to do with using `+=` over `<<`/`concat` (it's my understanding the latter is more performant in not creating additional objects while exhibiting the same resulting data structure), but I also don't think that should hold this PR up.

cbeer: I was just suggesting that, ideally, the whole thing will go away soon; I see it as just a hack to ensure complete parity with the old output. Since you seem concerned about it, I've replaced it with a single array without any concat'ing.

PR #850, opened 2017-11-07
use a special sort field for Page detail titles (accounting for paren…

### LINKED ISSUE #841: Provide different title_sort for Page details type
For title_sort, sort by Manuscript title : Page title

mejackreed:
…t in title)

Closes #841 

screen shot 2017-11-07 at 6 34 22 pm

### DIFF
--- a/lib/traject/macros/canvas.rb
+++ b/lib/traject/macros/canvas.rb
@@ -27,6 +27,16 @@ module Macros
       end
     end
 
+    ##
+    # Note: This method assumes an "enhanced" canvas with additional properties
+    # added beyond the IIIF Canvas model.
+    def extract_canvas_label_sort
+      lambda do |record, accumulator, _context|
+        labels = [record['manifest_label'].to_s, record['label'].to_s].reject(&:empty?)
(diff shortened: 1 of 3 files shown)


lib/traject/macros/canvas.rb
```diff
@@ -27,6 +27,16 @@ def extract_canvas_label
       end
     end
 
+    ##
+    # Note: This method assumes an "enhanced" canvas with additional properties
+    # added beyond the IIIF Canvas model.
+    def extract_canvas_label_sort
+      lambda do |record, accumulator, _context|
+        labels = [record['label'].to_s, record['manifest_label'].to_s].reject(&:empty?)
+        accumulator << labels.reverse.join(': ')
```
jkeck: Any particular reason we're reversing here and not ordering the array above?  I know it's the same array as in the other method, but just wondering if we can do w/o the reverse. /shrug

mejackreed: No reason. I changed it to order properly. Thanks for the suggestion.

Issue #846, opened 2017-11-07
Where should the source of truth be for a "Page detail" additional context fields?


mejackreed:
For the concatenated title and "identifier", should we get the data from the custom manifest, or from the mods?

We can do both quite easily, but it may be hard to keep up to date in all places.

@blalbrit

mejackreed:
Ok.. per @blalbrit 

Manuscript title: pull from manuscript manifest
Identifier: pull from manuscript mods/ document: `identifier_ssim`
Range: pull form manuscript manifest

PR #851, opened 2017-11-08
Index the identifier from the SolrDocument into the Canvas document.

### LINKED ISSUE #832: Add "identifier" to Page/Canvas/AnnoList for indexing (passed from parent)
Should this come from SDR object or from external manifest?

jkeck:
Closes #832 

## canvas-3f47f78ba92d14dd09ebe23bde585da0
canvas-3f47f78ba92d14dd09ebe23bde585da0

In this PR I simply just took the same array of notes and indexed them into the same field (this way they have the same behavior).  I think this may be undesirable behavior in that it is using Blacklight's default joining behavior (which causes values to be joined with comas and an "and").

A few possible approaches to deal with this:

1. Configure `identifier_ssim` to not be joined using Blacklight's default behavior.
    * This will affect any other documents with multiple identifiers, however maybe that's desirable for this kind of data?  Not sure.
1. Concatenate the identifiers for canvas/page detail documents into a single string.
    * This won't affect the identifier display for the manuscript itself however.
1. Index the identifier into a separate field for configuration.
    * This also won't affect the identifier display for the manuscript document itself.
1. Other ideas?

### DIFF
--- a/app/models/iiif_canvas_indexer.rb
+++ b/app/models/iiif_canvas_indexer.rb
@@ -26,6 +26,7 @@ class IiifCanvasIndexer
         # so we do this by enhancing the stored Hash with needed fields.
         enhanced_canvas = JSON.parse(canvas.to_json)
         enhanced_canvas['manifest_label'] = manifest.label
+        enhanced_canvas['parent_identifier'] = document.fetch('identifier_ssim', nil)
         canvas_resource.data = enhanced_canvas
         canvas_resource.save_and_index
       end
--- a/lib/traject/canvas_config.rb
+++ b/lib/traject/canvas_config.rb
@@ -16,6 +16,7 @@ end
 to_field 'id', extract_canvas_id
 to_field 'iiif_canvas_id_ssim', extract_canvas_iiif_id
 to_field 'format_main_ssim', literal('Page details')
+to_field 'identifier_ssim', extract_canvas_parent_identifier
 
 to_fields %w(title_display title_full_display title_uniform_search), extract_canvas_label
 to_field 'title_sort', extract_canvas_label_sort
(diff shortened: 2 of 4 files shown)


mejackreed:
I would be 👍  with the current approach with no changes. For the reason that it displays and is stored consistently with the identifier from the manuscript.

jkeck:
👍 I've removed the `WIP`.  I'll mention the display issue and standup and see if we should spin off a separate issue to deal w/ that.

blalbrit:
I'm okay with the default behavior here.

jkeck:
That's a good question.  According to https://github.com/sul-dlss/exhibits/issues/846#issuecomment-342608347 it should be the same content that is indexed into the `identifier_ssim` into the parent document.

If that's not the case, could you point me to where the source of this metadata is intended to be?

blalbrit:
Thanks @jkeck - I think the communication was that the identifier should come from the MODS (as opposed to the external manifest), not sure when `identifier_ssim` got included, as I don't know what gets indexed into that solr field (well... didn't know, now I do). What we would be looking for is the information that will be held in:

`<shelfLocator>MS 69</shelfLocator>`

In the `<location>` block. See https://purl.stanford.edu/cf386wt1778.mods for a fixture.

mejackreed:
So I'm now understanding @blalbrit that this is different than the mocks and maybe where that is where the confusion has come in. I was working off the assumption that we were implementing this:

32465663-78c23d66-c301-11e7-98d0-4245b08b79fb

I think what you are describing @blalbrit might be a bit different. Can we discuss briefly to see how to move forward.

blalbrit:
that totally makes sense @mejackreed - I should have caught it at the mock stage. Thanks all for patiently working through this with me.

PR #852, opened 2017-11-08
Update spotlight / Add "resource type" index/show field behind feature flag

### LINKED ISSUE #815: Update configuration to make "format_main_ssim" an index field
We currently show Resource type as a facet 

```
 config.add_facet_field 'format_main_ssim', label: 'Resource type', limit: true
```

but it's not configured as an index field and doesn't show up in the list of potential metadata fields.

I think we do want to show the Resource Type field in search results (and possibly the show pages), because it could be important for the user to be able to distinguish between the type of documents available, and I don't think we want to rely entirely on the thumbnail variations for distinguishing document types.

mejackreed:
Closes #815 

To resolve this we needed to update Spotlight. Spotlight recently introduced a change from FactoryGirl to FactoryBot so we also needed to make that change locally.

screen shot 2017-11-08 at 9 53 08 am

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -62,7 +62,7 @@ end
 
 group :test do
   gem 'rspec-rails'
-  gem 'factory_girl_rails'
+  gem 'factory_bot_rails'
   gem 'rails-controller-testing'
   gem 'capybara'
   gem 'chromedriver-helper'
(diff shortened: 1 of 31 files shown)


camillevilla:
@mejackreed I'm not sure this in scope for this PR, but what do you think about removing the `FactoryBot` prefix? Because we use `config.include FactoryBot::Syntax::Methods` in our `rails_helper`, the prefix isn't needed. I doubt FactoryBot is going to change names again but it might be a nice opportunity to DRY things out.

mejackreed:
Thanks for the suggestion @camillevilla . I've updated the PR accordingly.

PR #860, opened 2017-11-09
Index the canvas' parent manuscript number in place of the identifier.

### LINKED ISSUE #858: Remove identifier_ssim from Page details, Add "manuscript number" to Page detail objects

jkeck:
Closes #858

### canvas-dddcd264a38dea30849f17da8ddb59c2

fh878gz0315-canvas-dddcd264a38dea30849f17da8ddb59c2

### DIFF
--- a/spec/models/iiif_canvas_indexer_spec.rb
+++ b/spec/models/iiif_canvas_indexer_spec.rb
@@ -19,7 +19,7 @@ RSpec.describe IiifCanvasIndexer do
       id: druid,
       content_metadata_type_ssm: ['image'],
       iiif_manifest_url_ssi: 'http://example.com',
-      identifier_ssim: %w(ABC 123 xyz)
+      manuscript_number_tesim: %w(MS 123)
     )
   end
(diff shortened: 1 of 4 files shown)


spec/models/iiif_canvas_indexer_spec.rb
```diff
@@ -19,7 +19,7 @@
       id: druid,
       content_metadata_type_ssm: ['image'],
       iiif_manifest_url_ssi: 'http://example.com',
-      identifier_ssim: %w(ABC 123 xyz)
+      manuscript_number_tesim: %w(ABC 123 xyz)
```
camillevilla: @jkeck should this be changed to a single string, like e.g. `MS 079`? The test still works, but it might make it easier to scan this spec for expected behavior.

jkeck: 👌 It still has to be an array because it's a multi-valued field in solr, but I've added a more realistic value into that array.

PR #862, opened 2017-11-09
add manifest druid and @id url to canvas indexing and json - closes #849

### LINKED ISSUE #849: Attach a manifest link to the Canvas model.
- [x] choose content type for canvas resources
- [x] provide access to parent manuscript druid (fixed by #851)
- [x] build or pass custom manifest id through to canvas concern

aeschylus:


jkeck:
LGTM, I would like somebody else to review since we paired on most of this together.

app/models/iiif_canvas_indexer.rb
```diff
@@ -27,6 +27,7 @@ def index_canvases
         enhanced_canvas = JSON.parse(canvas.to_json)
         enhanced_canvas['manifest_label'] = manifest.label
         enhanced_canvas['parent_manuscript_number'] = document.fetch('manuscript_number_tesim', nil)
+        enhanced_canvas['parent_druid'] = document.id
```
mejackreed: Is this meant to represent the druid of the "manuscript" (SDR object) ? Is it similar to this? 

https://github.com/sul-dlss/exhibits/commit/84b57aaa49c7c129c8b1a14cf9720a61f3ccc110

jkeck: Ah, we didn't catch that in the resulting solr document.  Yeah, that's what it was intended to be, so we can probably just re-use that field.

aeschylus:
@mejackreed Should be fixed up now.

PR #863, opened 2017-11-10
Bugfix: isolate collapsible Text title / tableOfContents

### LINKED ISSUE #856: Text title collapse needs to use a unique id
When viewing the show / hide on a search results page, this becomes problematic:

showhide

camillevilla:
Closes #856 

This PR
- assigns unique IDs to each instance of collapsing tableOfContents
- fixes JS so that clicking on one `Show` / `Hide` link doesn't affect the others
- moves `.caret-right` class out of Parker theme

## Before
showhide

## After
fixed-table-of-contents

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -43,6 +43,7 @@ module ApplicationHelper
     return if options[:value].blank?
     contents = options[:value][0].split('--').map(&:strip)
     contents = contents.join('<br />').html_safe
-    render partial: 'catalog/table_of_contents', locals: { contents: contents }
+    id = @document.id
+    render partial: 'catalog/table_of_contents', locals: { contents: contents, collapse_id: "collapseToc-#{id}" }
   end
 end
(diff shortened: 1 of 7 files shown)


app/helpers/application_helper.rb
```diff
@@ -43,6 +43,7 @@ def table_of_contents_separator(options = {})
     return if options[:value].blank?
     contents = options[:value][0].split('--').map(&:strip)
     contents = contents.join('<br />').html_safe # rubocop:disable Rails/OutputSafety
-    render partial: 'catalog/table_of_contents', locals: { contents: contents }
+    id = options[:document][:id]
```
cbeer: Is this a real `SolrDocument` instance? If so, you probably want to call the `#id` method instead of using hash-access.

camillevilla: Thanks for the tip, @cbeer. I've updated my helper and spec accordingly.

PR #864, opened 2017-11-10
Remove unused DOR indexing configuration

cbeer:


lib/traject/dor_config.rb
```diff
@@ -20,26 +20,18 @@
   next if resource.collection?
```
cbeer: I think @jkeck said the `display_type` was effectively unused; should we remove it?

PR #868, opened 2017-11-10
Update dependencies

cbeer:

PR #870, opened 2017-11-10
Remove backticks, which were causing system execute calls.

jkeck:


mejackreed:
Did this commit do the same thing? https://github.com/sul-dlss/exhibits/pull/868/commits/2e7475b0f69ace6fef6f408b62f97fe90bccb92e

PR #871, opened 2017-11-10
single notes should not be wrapped in a ul/li

### LINKED ISSUE #865: Add conditional handling of single note versus multiple notes
We recently added code to put multiple note fields into a `ul`. This works fine when there is more than one note. When there is only one note, however, it looks a bit odd to have a single bulleted item:

stanford_residences_--_sacramento__-_stanford_historical_photograph_collection_-_spotlight_at_stanford

(The `ul` also adds a bit of extra whitespace below the element.)

If possible, I suggest we add some conditional code so that if there only one note, we present it in the `dd` without the wrapping `ul` so it displays similar to other metadata fields.

mejackreed:
Closes #865 

### Fixtures
`qr683hn3144`

Could not find a fixture w/ multiple notes though.. anyone know of one handy?

screen shot 2017-11-10 at 3 34 58 pm

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -25,6 +25,7 @@ module ApplicationHelper
 
   def notes_wrap(options = {})
     return if options[:value].blank?
+    return options[:value].first if options[:value].count == 1
     content_tag('ul', class: 'general-notes') do
       safe_join(options[:value].collect do |note|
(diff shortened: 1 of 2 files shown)


camillevilla:
@mejackreed `fixtures/book.bib` works. I can't think of an SDR object off the top of my head though.

PR #873, opened 2017-11-10
Update to Rails 5.1

cbeer:


mejackreed:
Have we checked all of the Rails / jQuery UJS things? I remember that being an upstream concern in Spotlight

cbeer:
I think it's fine here, since we were already on Turbolinks 5.x and using rails_ujs.

PR #875, opened 2017-11-11
Index IIIF range labels into canvas resources.

### LINKED ISSUE #834: Add section to CanvasResource indexing
~~Blocked by #833~~

Fixture now available in #833

jkeck:
Closes #834 

## canvas-8bf659aec67889770acce7c7442270a1 (druid: fh878gz0315)
screen shot 2017-11-13 at 8 25 47 am


app/models/iiif_manifest_harvester.rb
```diff
@@ -16,9 +18,26 @@ def canvases
     manifest.sequences.first.canvases
   end
 
+  def ranges_for(canvas_id)
+    ranges.select do |range|
+      normalized_canvases = range.canvases.map { |c| c.partition('#').first }
```
jkeck: I _think_ we're doing the correct thing here in that if the canvas ID stored in the range has a reference to a xywh section we would still want it to match the given canvas ID that we're indexing.

aeschylus: Yes, this is correct, for indexing purposes.

PR #880, opened 2017-11-13
Refactor DOR traject configuration to use existing macros

cbeer:


lib/traject/macros/general.rb
```diff
@@ -5,7 +5,7 @@ module General
     def accumulate(&lambda)
       lambda do |record, accumulator, context|
         Array(lambda.call(record, context)).each do |v|
```
camillevilla: @cbeer are you still considering renaming this lambda to distinguish it from `lambda do`?

cbeer: Yes, I'll rename all the lambda arguments for clarity.

camillevilla:
yep, LGTM ✅

PR #881, opened 2017-11-13
limit size of error messages when saving to database

### LINKED ISSUE #709: [exhibits/stage] ActiveRecord::StatementInvalid: Mysql2::Error: Got a packet bigger than 'max_allowed_packet'
## Backtrace

    line 6 of [PROJECT_ROOT]/app/jobs/record_index_status_job.rb: perform

[View full backtrace and more info at honeybadger.io](https://app.honeybadger.io/projects/49092/faults/34896431)

[exhibits/stage] ActiveRecord::StatementInvalid: Mysql2::Error: Got a packet bigger than 'max_allowed_packet' bytes: UPDATE `spotlight_solr_document_sidecars` SET `index_status` = x'2d2d2d0a3a6f6b3a2066616c73650a3a6d6573736167653a207c2d0a2020233c52536f6c723a3a4572726f723a3a487474703a2052536f6c723a3a4572726f723a3a48747470202d203430302042616420526571756573740a20204572726f723a207b22726573706f6e7365486561646572223a7b22737461747573223a3430302c225154696d65223a313732347d2c226572726f72223a7b226d65746164617461223...

drh-stanford:
This PR fixes #709, or at least fixes the buffer overflow issue.

### DIFF
--- a/app/models/dor_harvester.rb
+++ b/app/models/dor_harvester.rb
@@ -63,8 +63,7 @@ class DorHarvester < Spotlight::Resource
                 exception_or_message.inspect
               else
                 exception_or_message.to_s
-              end
-
+              end.truncate(1.megabyte)
     RecordIndexStatusJob.perform_later(self, resource.bare_druid, ok: false, message: message)
   end
 
--- a/spec/models/dor_harvester_spec.rb
+++ b/spec/models/dor_harvester_spec.rb
@@ -77,6 +77,14 @@ describe DorHarvester do
       it 'records an indexing error for a druid' do
         expect { subject.on_error(resource, 'error message') }.to have_enqueued_job(RecordIndexStatusJob).with(harvester, druid, ok: false, message: 'error message')
       end
+      it 'records an indexing exception for a druid' do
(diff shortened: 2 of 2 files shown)


app/models/dor_harvester.rb
```diff
@@ -1,3 +1,5 @@
+include ActionView::Helpers::TextHelper
```
cbeer: 👎 to including this helper globally.

app/models/dor_harvester.rb
```diff
@@ -64,8 +66,10 @@ def on_error(resource, exception_or_message)
               else
                 exception_or_message.to_s
               end
-
-    RecordIndexStatusJob.perform_later(self, resource.bare_druid, ok: false, message: message)
+    RecordIndexStatusJob.perform_later(self,
+                                       resource.bare_druid,
+                                       ok: false,
+                                       message: truncate(message, limit: 1.megabyte))
```
cbeer: Actually, Rails makes `String#truncate` available, so we shouldn't need `TextHelper` at all.

drh-stanford:
Great, thanks for the feedback. I didn't know about `String#truncate`. I've updated the PR

PR #844, opened 2017-11-07
add manifest label/title to page detail title

mejackreed:
Closes #831 

This PR enhances a  canvas object that is passed into the traject indexer with the "manifest_label" AKA title. This is needed to create the "Page detail" title outlined in designs for Page detail search and show pages.

Fixture object: `bg021sq9590`

screen shot 2017-11-07 at 2 06 02 pm
screen shot 2017-11-07 at 2 05 54 pm

PR #890, opened 2017-11-14
Metadata show partial + basic default mods rendering

### LINKED ISSUE #883: Add a button to display a modal that shows html mods (adding a route to support non-JS)
Design in #799 

mock

 - [x] Button on show page that opens up modal with mods xml display page
 - [x] Button is a link to mods xml display page with no js

See Blacklight `add_show_tools_partial`
https://github.com/projectblacklight/blacklight/wiki/Adding-new-document-actions

mejackreed:
Closes #883 
Also updates Blacklight to get a JS fix we needed.
Fixture.. anything :)

This only feature flags the button but not the route (I think thats ok :) )

mdbutton

## UPDATE!!

Added the basic metadata html

mdstuffs

Addresses #844.

I think additional changes / designs should be new issues.


app/views/catalog/_metadata_button_default.html.erb
```diff
@@ -0,0 +1,14 @@
+<% if feature_flags.view_all_metadata? %>
```
cbeer: Now that we have the MODS display stuff working, do we need the feature flag still?

mejackreed: Great idea. I've pulled that out now.

config/routes.rb
```diff
@@ -28,11 +28,20 @@
...
   mount Spotlight::Engine, at: '/'
+  concern :exportable, Blacklight::Routes::Exportable.new
 
   resources :exhibits, path: '/', only: [] do
     resource :dor_harvester, controller: :"dor_harvester", only: [:create, :update]
     resource :bibliography_resources, only: [:create, :update]
     resource :viewers, only: [:create, :edit, :update]
+
+    resources :solr_documents, except: [:index], path: '/catalog', controller: 'catalog' do
+      concerns :exportable
```
cbeer: I'm a little surprised we need to re-mix-in `exportable`. What do you see without it?

config/routes.rb
```diff
@@ -28,11 +28,20 @@
...
 
   mount Spotlight::Engine, at: '/'
+  concern :exportable, Blacklight::Routes::Exportable.new
 
   resources :exhibits, path: '/', only: [] do
     resource :dor_harvester, controller: :"dor_harvester", only: [:create, :update]
     resource :bibliography_resources, only: [:create, :update]
     resource :viewers, only: [:create, :edit, :update]
+
+    resources :solr_documents, except: [:index], path: '/catalog', controller: 'catalog' do
```
cbeer: Instead of overriding the normal routes we're getting from spotlight, would `only: []` work (instead of `except: [:index]`)?

mejackreed: Yep! Added.. thanks

config/routes.rb
mejackreed: `No route matches..` and `undefined method `exhibit_root_path' for #<Spotlight::CatalogController:0x007ff41d7be2b0> Did you mean? exhibit_viewers_path`

PR #897, opened 2017-11-15
only show view all metadata when mods-xml is present

mejackreed:
This makes sure the link doesn't show up on pages like canvases and bib items

### DIFF
--- a/app/views/catalog/_metadata_button_default.html.erb
+++ b/app/views/catalog/_metadata_button_default.html.erb
@@ -1,12 +1,14 @@
-<div class='container'>
-  <div class='row'>
-    <% # Using the Bootstrap container/row/col structure here to easily get this on its own line %>
-    <div class='col-md-12'>
-      <%= link_to 'View all metadata »',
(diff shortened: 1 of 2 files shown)


app/views/catalog/_metadata_button_default.html.erb
```diff
@@ -1,12 +1,14 @@
-<div class='container'>
-  <div class='row'>
-    <% # Using the Bootstrap container/row/col structure here to easily get this on its own line %>
-    <div class='col-md-12'>
-      <%= link_to 'View all metadata »',
-                  metadata_exhibit_solr_document_path(current_exhibit, document),
-                  data: { ajax_modal: 'trigger' },
-                  class: 'btn btn-default'
-      %>
+<% if @document.fetch(:modsxml, nil) %>
```
cbeer: Ought we add a `#modsxml?` method or similar to `SolrDocument`?

mejackreed: Added. :)

PR #902, opened 2017-11-16
Scroll the metadata body in the modal

### LINKED ISSUE #901: Metadata modal height should be a fixed height with a vertical scroll body.
> My interpretation of Gary's mockup is that the popup is a fixed height within the viewport, while the modal-body div is scrollable. Close and Download buttons are intended to be always visible / fixed at the bottom.

From: https://github.com/sul-dlss/exhibits/issues/799#issuecomment-344775935

See: https://www.bootply.com/WaTI0dBpa3

mejackreed:
Closes #901 

scrolly

### DIFF
--- a/app/assets/stylesheets/modules/view_metadata.scss
+++ b/app/assets/stylesheets/modules/view_metadata.scss
@@ -13,4 +13,8 @@
       display: inline-block;
     }
   }
+  .modal-body {
+    max-height: calc(100vh - $modal-header-footer-height-guess);
+    overflow-y: auto;
+  }
 }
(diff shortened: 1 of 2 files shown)


app/assets/stylesheets/modules/view_metadata.scss
```diff
@@ -13,4 +13,8 @@
       display: inline-block;
     }
   }
+  .modal-body {
+    max-height: calc(100vh - 212px);
```
cbeer: Where does `212px` come from? Height of the title + height of the footer? Maybe it could get pulled out to a scss variable to remove uncertainty, or at least get a comment?

mejackreed: Just a guess of what the header + footer will be. I've pushed it in to a variable. I don't think we can know what the actual calculation will be.

Issue #548, opened 2017-09-18
Support date ranges in display and facets


caaster:
Story:
Metadata for special collections materials often provides date values as approximate ranges. Exhibit curators would like to provide an accurate representation of this information for users, with functionality on par with SearchWorks.

Acceptance criteria:
- [x]  The date display field should represent the values of /mods/originInfo/date*[@point=”start”] and /mods/originInfo/date*[@point=”end”] (where date* is dateCreated, dateIssued, or dateCaptured) as a range, and include the value of the qualifier attribute when present. 
 - [ ] A date slider facet (#994, #971) that works for date ranges in metadata should be available.

blalbrit:
For Parker, see: https://purl.stanford.edu/cf386wt1778 for instance.

caaster:
A researcher using the Herbert Matter exhibit has reported their confusion that many of the image descriptions are annotated: "[1937]" -- which is actually incorrect for many images. For accuracy/better research it would be helpful for both the approximate start and end dates to be displayed. Here is an example from the exhibit: https://exhibits.stanford.edu/herbert-matter/catalog/rx795tx1655

arcadiafalcone:
To expand on this --
What displays in Spotlight:
1937
What should display (by mods_display rules):
[ca. 1937-1984]

Often with archives if an individual object is undated, the date span of the collection as a whole will be applied to the object to give at least a rough idea of the time period. Omitting the date range and qualifier (#551) creates a significant difference between the information in the metadata and what is displayed to the user.

mejackreed:
Thanks for the help @arcadiafalcone and @caaster ! We have implemented the display in #906, looking into the date range facet slider seems to bring up more questions for me.

It seems like in SearchWorks (for SDR content) we index a single year. See: https://github.com/sul-dlss/sw-indexer/blob/master/lib/sw_mapper.rb#L102

We can do this same behavior in Exhibits, but it doesn't solve the range issue. Should we move ahead without the range information (and just use StanfordMods `pub_year_int` logic) ? If not I think we need more information about how StanfordMods should parse and understand an indexed date range.

arcadiafalcone:
I hadn't realized the date slider facet in SearchWorks doesn't handle MODS date ranges. I would rather update the slider to have range functionality before implementing it in Spotlight. Having the slider reference only a single date causes relevant objects to disappear from facet results, and does so in a way that is completely invisible to the user, unless they already know the content well enough to realize that something is missing.

PR #906, opened 2017-11-16
Update date indexing and display

### LINKED ISSUE #551: Support date qualifiers in display
Story:
Dates are sometimes given as approximations using the “qualifier” attribute. Exhibit curators would like to provide an accurate representation of date uncertainty for users.

Acceptance criteria:
- Single date display should include qualifier information. (This reproduces current SearchWorks behavior.)
- Date range display should include qualifier information. (This reproduces current SearchWorks behavior.)

cbeer:
Fixes #551 
Partially address #548 

- Uses the new stanford-mods imprint logic
- Add a new date field for display that includes qualifiers and ranges
- Fixes incorrect indexing of `1uuu` dates


lib/traject/dor_config.rb
```diff
@@ -79,6 +79,19 @@
...
 to_field 'pub_year_w_approx_isi', stanford_mods(:pub_year_int, false)
+
+to_field 'date_display' do |resource, accumulator, _context|
+  imprint = ModsDisplay::Imprint.new(nil, ModsDisplay::Configuration::Imprint.new, nil)
+
+  Array(resource.smods_rec.origin_info).map do |value|
+    dates = imprint.dates(value)
+    accumulator.concat(dates.map(&:values).flatten)
+
+    part = imprint.send(:parts_element, value)
```
drh-stanford: is this calling a private method in the Imprint class? should we rework mods_display to make that a public method?

cbeer: Yes, we should. It's going to be an unfortunate amount of work, though, to refactor that code and reconcile it with stanford-mods (which seems to have a copy/pasted version of the code, which was later tweaked).

camillevilla:
@cbeer I'm having difficulty getting this to work locally. I tried adding the two druids provided in #551 and #548 (tw368kj5707, rx795tx1655). Item status shows they both throw the following error: 
```
#<RSolr::Error::Http: RSolr::Error::Http - 400 Bad Request Error: {"responseHeader":{"status":400,"QTime":1},"error":{"metadata":["error-class","org.apache.solr.common.SolrException","root-error-class","org.apache.solr.common.SolrException"],"msg":"ERROR: [doc=rx795tx1655] unknown field 'date_display'","code":400}} 
...
```

Not sure if this is a problem with my dev environment, but I would appreciate someone else giving this a test drive.

mejackreed:
@camillevilla I've added a commit that adds the needed field in the solr schema.xml. I've also created a PR to sul-solr-configs https://github.com/sul-dlss/sul-solr-configs/pull/107 that when merged by devops will update our prod/stage config.

cbeer:
Ugh. Sorry about that. I wonder if we want to switch from `_display` to one of the `*_ssim` fields?

mejackreed:
@cbeer Yeah I'm good with that. I'll update the commit now and close the sul-solr pr's.

camillevilla:
👍 works fine for me now

Issue #910, opened 2017-11-17
Ordering & grouping of metadata


jvine:
After analysis discussion with gryphon-search and @arcadiafalcone: 
- the grouping and ordering in SearchWorks (and by default applied to PURL) was done to support specific goals and tasks on the show page that do not apply in the same way to either PURL or Exhibits in general, or Parker in particular
- there is a general feeling that a consistent grouping of MODS might be desirable, but will require substantially more analysis before it can be applied broadly across Exhibits (what seems to work for Parker doesn't fit all metadata equally well) or elsewhere
- for now, I am proposing that the metadata be displayed in its standard order, with some exceptions/adjustments (listed below).

---
- [x] 1. Apply SW/PURL visual design to the `<dl>` elements. (Transform `<dt>` to uppercase; indent `<dl>`; adjust line-spacing.) #918

- [x] 2. Where the title exists as a heading on the page, do not include it in the labelled metadata. Split out to #922

- [x] 3. Use `<roleTerm>` as labels in place of "Contributor" or "Author/Creator" labels. Split out to #936 

- [x] 4. Convert the "Contents" display from '--' delimiters to a bulleted list (without show/hide behaviour in the context of the modal):
  - [x] Change `--` into `<li>` list (See https://github.com/sul-dlss/mods_display/pull/42)
  - [x] Suppress Contents in Parker exhibits closed in #944

- [x] 5. Separate the rights & access metadata into a separate section with a heading of "Access conditions" (follows PURL and SW) (See #920)

- [x] 6. Display nested metadata (currently not displayed at all by mods_display) with show/hide behaviour (see #903 and #905). Note that the types (constituent and host) are labelled separately.

7. Consider displaying curator-added exhibit tags in the full metadata display, between the main metadata and Access conditions (not Parker-specific).

---

![mods_display1](https://user-images.githubusercontent.com/6945691/32923850-a02956c4-caee-11e7-924d-bfc18aaf1b50.jpg)

![mods_display2](https://user-images.githubusercontent.com/6945691/32923862-a65427a4-caee-11e7-9006-8fe77568045f.jpg)

jvine:
@drh-stanford based on a separate discussion, i think we intend to suppress the Table of Contents (i.e. item 4 above) from the metadata view, and include it only on the Show page. I'm not sure whether this is a Parker-specific change; I'll verify. Let me know if you have questions or this needs a separate issue.

mejackreed:
[...]

PR #920, opened 2017-11-20
Separate rights

mejackreed:
Partial implementation of #910 

This PR also removes the module mixin in CatalogController in favor of a pattern from Purl where the accessor is a PORO.

### Fixture hq225cp3879

rights


camillevilla:
@mejackreed, what's a PORO? ._.

mejackreed:
Ah :) Plain Old Ruby Object == PORO... Some historical reading :)  http://blog.steveklabnik.com/posts/2011-09-06-the-secret-to-rails-oo-design

jkeck:
👍 this won't allow us to use controller context, but I'm not sure that is needed at this point.

PR #927, opened 2017-11-20
hide Parker specific search fields under a feature flag

### LINKED ISSUE #879: Implement: hide Parker-specific search fields behind a feature flag
See #777 for investigation work

Parker-specific fields are currently disabled be default.

mejackreed:
Closes #879

### DIFF
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -15,6 +15,7 @@ feature_flags:
   bibliography_resource: false
   index_related_content: false # A feature that will index related SDR content. Initial implemented for Parker's IIIF AnnotationLists
   add_resource_type_index_field: false # Should the resource type field be added as an index/show field. For some exhibits this is desirable
+  add_parker_search_fields: false # Parker specific search fields
 traject:
   processing_thread_pool: 1
 iiif_dnd_base_url: https://library.stanford.edu/projects/international-image-interoperability-framework/viewers?%{query}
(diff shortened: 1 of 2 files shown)


mejackreed:
I'm not really sure if Coveralls is correct here?

config/settings.yml
```diff
@@ -15,6 +15,7 @@ feature_flags:
   bibliography_resource: false
   index_related_content: false # A feature that will index related SDR content. Initial implemented for Parker's IIIF AnnotationLists
   add_resource_type_index_field: false # Should the resource type field be added as an index/show field. For some exhibits this is desirable
+  add_parker_search_fields: false # Parker specific search fields
```
jkeck: Just trying to think through the implications of this and wondering a) is this a flag that we will intend to remove at some point, and b) should we use an exhibit specific flag for this?

mejackreed: Hmm. I'll try and answer your questions here:

a) Frankly I don't know this answer. I think this will likely set a pattern if there are any other super custom search fields that need to be added this pattern could used. Do you have a suggestion?
b) I thought I was kind of doing it that way... a global false, followed by shared_configs PR's that whitelist parker exhibits (already submitted)

jkeck: I guess I kinda worry about `a` being abused as a pattern (don't have any suggestions otherwise that we can implement w/i our timeframe).

👍 on `b` I hadn't seen that PR.

mejackreed: Sure understood. Any suggestions about mitigating `A` ?

jkeck: Sadly no.  😢

PR #930, opened 2017-11-21
setup viewer to display default on feature pages, but configured vers…

### LINKED ISSUE #767: When using Mirador as viewer, there is some widget weirdness happening
### Actionable work:
 - [ ] for item embeds, use the default viewer not the configured one.

See:

screenshot 2017-10-18 16 12 13

mejackreed:
…ion on show pages

Closes #767 

I'm unsure how to actually test this, as my SirTrevor mocking did not seem to work... Halp

viewerconfig

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -54,4 +54,17 @@ module ApplicationHelper
     title = title.include?(':') ? title.partition(':')[2] : ms_number[0]
     link_to title, spotlight.exhibit_solr_document_path(current_exhibit, druid[0])
   end
+
+  ##
+  # Renders a viewer for an object with understanding of the context. In the
+  # context of spotlight/catalog render the configured viewer. In other contexts
+  # (feature page) render the default viewer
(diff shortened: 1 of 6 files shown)


jkeck:
What did you try wrt sir-trevor widget mocking?  Can we make it a js spec that can add the widget to the page and validate the correct viewer's markup is present?

mejackreed:
The things I tried mostly involved trying to directly add the `content` into the feature page https://github.com/sul-dlss/exhibits/pull/930/files#diff-abf5e8f2b2bbf8c6bfd76166c0ad0e39R78 or unit testing the helper method. Unit testing the helper method got hairy because I had to mock a bunch of understanding from the controller.

I'll work on the js approach.

camillevilla:
Yep! This works. Adding a before screenshot so folks don't have to dig through the discussion in #767. @mejackreed did you get any further with the JS tests?

## Before
mirador_embed

## After
default_embed

mejackreed:
Nope I couldn't get any further with the feature tests 😢

PR #931, opened 2017-11-21
Remove extra title in metadata modal

### LINKED ISSUE #922: Where the title exists as a heading on the page, do not include it in the labelled metadata.
discussion in #910

camillevilla:
Closes #922 

This PR removes the `Title` field from the labeled metadata. The title is already displayed in the modal as an `<h3>`. 

## Before
showing extra title in metadata list

## After
hiding extra title in metadata list

### DIFF
--- a/app/views/catalog/_metadata.html.erb
+++ b/app/views/catalog/_metadata.html.erb
@@ -1,3 +1,3 @@
-<%= @document.mods.to_html.html_safe %>
+<%= @document.mods.body.html_safe %>
 
 <%= render 'metadata_rights_access' %>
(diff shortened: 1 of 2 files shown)


drh-stanford:
I think (haven't tried it) you can do this in the Rails model rather than in CSS, like so in `ModsDisplayObject`:

```
  configure_mods_display do
    title do
      ignore!
    end
  end
```

camillevilla:
@drh-stanford I'll give that a try.

camillevilla:
Hmm. @drh-stanford looks like
```ruby
  configure_mods_display do
    title do
      ignore!
    end
  end
```
DOES work, but it also removes our alternative titles. I'm looking into how to drill down from `title`.

drh-stanford:
Ok, your CSS solution may be the best route then.

PR #938, opened 2017-11-23
Toggle nested related items

### LINKED ISSUE #905: Show/hide nested metadata for related items
Related to #903 for related items with nested metadata:

1. nested metadata is hidden by default
2. title is focusable/clickable, with an arrow icon in front

> related items closed

3. when user clicks a title, the nested metadata for that item opens in an indented `<dl>`. Arrow rotates 90 degrees CW when the nested section is open; clicking again closes the section and reverts the arrow.

> related item open

4. when there is more than one related item with nested metadata, the main label for the group gets an "expand all" link. 
     - Clicking the link opens all the nested items in the group; 
     - link text changes to "collapse all"
     - all arrows rotate 90 degrees CW
     - Clicking "collapse all" closes all nested items; label changes back to "expand all"; all arrows revert to starting state.

jkeck:
Closes #905 

TODO: 
- [x] Add feature spec (need to add a new or modify an existing fixture)
- [ ] ~~Add an expand/collapse all toggle~~ (broken out into #947)
- [x] Update changelog

## fh878gz0315
nested-related-items


app/assets/javascripts/nested_related_items.js
```diff
@@ -0,0 +1,61 @@
...
+      var listItem = item.parent('li');
+      listItem.toggleClass('open');
+      listItem.find('dl').toggle();
+    }
+  };
+
+  global.NestedRelatedItems = NestedRelatedItems;
+}(this));
+
+Blacklight.onLoad(function () {
```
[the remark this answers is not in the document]
jkeck: Added this as a `global` at the top.  Hound is no longer complaining.

app/assets/javascripts/nested_related_items.js
```diff
@@ -0,0 +1,63 @@
...
+      return list.find(this.options.itemSelector);
+    },
+
+    addToggleLink: function(content) {
+      var dl = content.find('dl');
+
+      if(dl.length > 0) {
+        // Hide and remove the dl (so we can deal with just the text below)
+        dl.hide();
+        dl.remove();
```
mejackreed: Have you considered using  a `.detach` here? https://api.jquery.com/detach/

> The .detach() method is the same as .remove(), except that .detach() keeps all jQuery data associated with the removed elements. This method is useful when removed elements are to be reinserted into the DOM at a later time.

I'm not sure if we need to keep that info around or not.

jkeck: 🤔 I'm not sure any jQuery will be associated with the `dl`, but happy to use `.detach` instead just in case.

mejackreed: /shrug 👍

jkeck:
👍 updated @mejackreed.

Another thought I had is I can replace the first commit with an update to a released version of ModsDisplay (happy to just do that as a separate PR too).

PR #946, opened 2017-11-27
hide bibtex partial behind feature flag

mejackreed:
This hides the ability for non-configured exhibits to index bibtex files, but unfortunately still shows the "pill" link.

disablebib

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ### Added
 ### Changed
+- BibTeX upload form is now only shown for Exhibits in which it is explicitly enabled through a feature flag #946
 ### Deprecated
 ### Removed
 ### Fixed
--- a/app/views/bibliography_resources/_form.html.erb
+++ b/app/views/bibliography_resources/_form.html.erb
@@ -8,4 +8,6 @@
       </div>
     </div>
   <% end if can? :manage, @resource %>
+<% else %>
+  This feature is not currently available for this exhibit.
 <% end %>
--- a/app/views/bibliography_resources/_form.html.erb
+++ b/app/views/bibliography_resources/_form.html.erb
@@ -1,9 +1,11 @@
-<%= bootstrap_form_for([current_exhibit, @resource.becomes(BibliographyResource)], layout: :horizontal, label_col: 'col-md-2', control_col: 'col-md-6', as: :resource) do |f| %>
-  <%= f.file_field :bibtex_file, label: 'BibTeX File', class: 'form-control', control_col: 'col-sm-6', required: true, accept: 'application/x-bibtex', help: 'Each BibTeX file entry will be added as an exhibit item with a “Reference” resource type.'  %>
-  <div class="form-actions">
-    <div class="primary-actions">
-      <%= cancel_link @resource, :back, class: 'btn btn-default' %>
-      <%= f.submit t('.add_item'), class: 'btn btn-primary' %>
+<% if feature_flags.bibliography_resource? %>
+  <%= bootstrap_form_for([current_exhibit, @resource.becomes(BibliographyResource)], layout: :horizontal, label_col: 'col-md-2', control_col: 'col-md-6', as: :resource) do |f| %>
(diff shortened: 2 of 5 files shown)


camillevilla:
re:pill showing ... how's about another Parker theme hack?

camillevilla:
Actually...hmm it does seems to be hiding the BibTeX pill link. Or were you talking about the SDR link? I did have to restart my rails server to see the difference.
## bibliography_resource: true
screen shot 2017-11-27 at 3 11 42 pm

## bibliography_resource: false
screen shot 2017-11-27 at 3 10 22 pm

mejackreed:
I'm seeing something a bit different.. were you restarting your rails app? this is strange.. CHANGELOG updated!

mejackreed:
Yeah, and now i remember that @caaster said she was telling exhibit curators not to use this rather than to keep this behind a feature flag.

jkeck:
Just took this for a spin and was able to hide the form by setting the `bibliography_resource` to false in my `development.local.yml`.

This is the default setting in production, and it appears that we set this to `true` in `shared_configs`, so this looks 👌 to me.

jkeck:
It seems like we need to make the css change still to remove the button in order to :shipit: ?

mejackreed:
I'm not sure how that would exactly work... and if we make the css change, then we don't really need the featureflag thing, right?

Re: CSS change

display:none for default theme, display:block parker theme?

jkeck:
Ah, fair point.  So what's the experience we want here? 

1. Button w/ form that we tell users not to use (what we currently have).
1. Button that when clicked shows an empty form (what this PR does).
1. Button only shows up for the Parker exhibit (via CSS).

mejackreed:
Maybe another option:

4. Just let people use this new feature

ggeisler:
> 4. Just let people use this new feature

I'm unclear if the indexing code that handles the BibTex upload is Parker-specific, but if the indexing will work for a curator in a non-Parker exhibit who uploads a valid BibTex file, this seems like the right approach to me.

mejackreed:
The indexing will work for any BibTeX file. There are some caveats only if the user needs to related the items to an SDR or other object.

ggeisler:
We (service team) could add some documentation for this feature to explain what it's for and how to take full advantage of it. But if the indexing works the documentation might not be a blocker to releasing the feature generally (i.e., odds seem low that many curators will try to use it in the near future). @caaster might have opinions from a service perspective.

camillevilla:
Re: BibTex stuff

There are two feature of our indexing that would make it harder to generalize:
- We assume that Zotero tags contain druids and index those into the `related_document_id_ssim` field. 
- the "Reference type" is derived from a small subset of BibTex / Zotero fields.

BibTeX type | Zotero type
------------ | -------------
phdthesis | Thesis
incollection | Book section
article | Journal article
book | Book
misc | Document

BibTeX indexing currently fails silently, and I think it would be frustrating to an end user. For example, I just tried to upload a Zotero Artwork citation and I got

I think further research would be necessary to understand how well we cover available Zotero fields. From a service perspective, this would mean telling users:
- only the 5 types of Zotero formats above are available
- if their citation changes, they'll need upload the `.bib` file again (reindexing won't work).

ggeisler:
Thanks the additional info @camillevilla. The "failing silently" aspect of this is concerning and I agree that is not an something we'd want users to experience. And while the five supported reference types seem like the ones mostly likely to exist in an uploaded BibTex file, we'd probably want that to be clear to a curator upfront. 

So maybe there is both documentation and development work (error reporting) necessary before making this feature more broadly available?

If others agree, do we go with:

> 3. Button only shows up for the Parker exhibit (via CSS).

and just make a ticket for future work to make this feature more robust and documented? That ticket would include undoing the CSS needed for option 3.

Option 2:

> 2. Button that when clicked shows an empty form (what this PR does).

would also be okay with me if we added a message to the empty form that displays, saying "This feature is not currently available for this exhibit." That way the user who clicks on the BibTex button at least knows what's going on.

mejackreed:
Hi all.. I've opted for @ggeisler 's modifed #2 proposal with a message. Please see updated gif at the top comment.

ggeisler:
@mejackreed Sounds good, although I'm not sure the gif is working correctly (at least for me, I don't see any _clicking on the button and showing the form message_ action).

mejackreed:
Updated sorry @ggeisler

jkeck:
This LGTM 👍 

Can we add a quick integration test for this under the `test-flag-exhibit-slug` flag just to validate the behavior?

mejackreed:
@jkeck test added.

PR #957, opened 2017-11-28
Use explicit [5.0] for all db migrations

drh-stanford:
Noticed this when I tried to rebuild my development database. I was getting errors without adding these qualifiers. I don't get why Travis doesn't complain about this...

```
$ bundle exec rails --version
Rails 5.1.4
$ bundle exec rails db:migrate
...
StandardError: Directly inheriting from ActiveRecord::Migration is not supported. Please specify the Rails release the migration was written for:

  class AddDocumentIndexToSolrDocumentSidecar < ActiveRecord::Migration[4.2]
```

I did find this in `spec/rails_helper.rb` -- https://github.com/sul-dlss/exhibits/blob/master/spec/rails_helper.rb#L36-L38

```
# Checks for pending migrations before tests are run.
# If you are not using ActiveRecord, you can remove this line.
ActiveRecord::Migration.maintain_test_schema! unless ENV['TRAVIS']
```

### DIFF
--- a/db/migrate/20170201233151_add_document_index_to_solr_document_sidecar.spotlight.rb
+++ b/db/migrate/20170201233151_add_document_index_to_solr_document_sidecar.spotlight.rb
@@ -1,5 +1,5 @@
 # This migration comes from spotlight (originally 20160929180534)
-class AddDocumentIndexToSolrDocumentSidecar < ActiveRecord::Migration
+class AddDocumentIndexToSolrDocumentSidecar < ActiveRecord::Migration[5.0]
   def change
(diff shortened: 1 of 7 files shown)


aeschylus:
I don't know what effects, if any, this has, but going by http://blog.bigbinary.com/2016/03/01/migrations-are-versioned-in-rails-5.html, this seems like a good idea.

drh-stanford:
I didn't notice any substanative changes to schema.rb -- just whitespace I thought. I'll take another look...

drh-stanford:
Ok, I did find one substantive change -- the addition of a couple indexes in `spotlight_reindexing_log_entries`, so I've checked in the db/schema.rb. Thanks!

PR #960, opened 2017-11-29
Add toggle all link to nested related items

### LINKED ISSUE #947: Add expand/collapse behavior for related items
Blocked by #905/#938
Related to #905 for related items with nested metadata:

When there is more than one related item with nested metadata, the main label for the group gets an "expand all" link. 
  - Clicking the link opens all the nested items in the group; 
  - link text changes to "collapse all"
  - all arrows rotate 90 degrees CW
  - Clicking "collapse all" closes all nested items; label changes back to "expand all"; all arrows revert to starting state.

related item open

camillevilla:
Closes #947

This PR adds an "Expand / Collapse all" link to nested related items in the metadata display.

Todo: 
- [x]  update specs
- [x] appease 🐺 and 👮
- [x]  update Changelog

## Before
nested-related-items

## After
toggle-all

### DIFF
--- a/app/assets/javascripts/nested_related_items.js
+++ b/app/assets/javascripts/nested_related_items.js
@@ -14,18 +14,19 @@
         _this.addToggleLink($(this));
       });
 
-      _this.addToggleAll();
-      _this.toggleAll();
+      _this.addToggleAll(el);
+      _this.toggleAll(el);
     },
 
     listItems: function(list) {
       return list.find(this.options.itemSelector);
     },
 
-    addToggleAll: function() {
-      var title = $('dt[title="Contains"]');
-      $(title).append('<a href="javascript:;" id="toggleAll" class="mods_display_related_item_label">Expand all</a>');
-      $(title).attr('id', 'mods_display_related_item_dt');
+    addToggleAll: function(content) {
+      // find the preceding <dt>
+      var title = $(content).parents('dd').prev();
+      title.append('<a href="javascript:;" id="toggleAll" class="mods_display_related_item_label">Expand all</a>');
+      title.attr('id', 'mods_display_related_item_dt');
     },
 
     addToggleLink: function(content) {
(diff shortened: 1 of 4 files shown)


app/assets/javascripts/nested_related_items.js
```diff
@@ -13,12 +13,21 @@
...
+      _this.addToggleAll();
+      _this.toggleAll();
     },
 
     listItems: function(list) {
       return list.find(this.options.itemSelector);
     },
 
+    addToggleAll: function() {
+      var title = $('dt[title="Contains"]');
```
jkeck: I get kinda nervous about using the `title` attr in the selector here.  It could have its `dt` value changed by a `displayLabel` and we would lose this behavior.  If we pass the `el` variable when we call this from the `init` function we might be able to traverse the `dt`/`dd` structure to find the `dt` instead of relying on the label value itself.

app/assets/javascripts/nested_related_items.js
```diff
@@ -13,12 +13,21 @@
...
+      $(title).append('<a href="javascript:;" id="toggleAll" class="mods_display_related_item_label">Expand all</a>');
+      $(title).attr('id', 'mods_display_related_item_dt');
```
jkeck: Can we just call `.append` and `.attr` directly on `title` instead of wrapping it in `$()`?

app/assets/stylesheets/modules/related_item.scss
```diff
@@ -1,19 +1,32 @@
+@mixin toggle-arrow {
```
mejackreed: nice

app/assets/javascripts/nested_related_items.js
```diff
@@ -44,6 +53,27 @@
...
     },
 
+    toggleAll: function() {
+      var _this = this;
+
+      $('#toggleAll').on('click', function(){
+        var toggleLink  = $(this);
+        toggleLink.toggleClass('open');
+        toggleLink.text(($(this).text() == 'Collapse all') ? 'Expand all' : 'Collapse all');
+        var links =  $('.mods_display_nested_related_item');
```
jkeck: Similarly, if we pass `el` when this is called in `init` we can use the `listItems()` function (which uses this same selector which is stored in the `options` object).

app/assets/javascripts/nested_related_items.js
camillevilla: ✅  Makes sense. I switched over to finding the `<dt>` immediately preceding the list's parent `<dd>`.

app/assets/javascripts/nested_related_items.js
camillevilla: 👍

app/assets/javascripts/nested_related_items.js
camillevilla: ✅  Alright, I *think* I did the thing.

app/assets/stylesheets/modules/related_item.scss
camillevilla: 🗑 🐼 ✨

PR #970, opened 2017-12-01
Enables page selection for SDR object embeds in widget

### LINKED ISSUE #536: Ability to select page to display for multi-page item in UV doesn't work (didn't work with sul-embed either)
As an exhibit creator, I have created a feature page for an exhibit: https://exhibits.stanford.edu/paci/feature/reaching-out-to-chinese

On the feature page, I am using several widgets. I am using the item embed widget to display druid:kq228ws6597 -- which is a multi-page object. I want page 3 to display as the default, so I am using the "change" option afforded to select page 3. Although in the widget/edit mode, it appears that my selection of page 3 as the default has been chosen (even when I hit "save") -- unfortunately page 1 still displays.

### LINKED ISSUE #827: Change UI for Embed widget to reflect current functionality (no longer can select page to display from within a multi-page object)
See #536

mejackreed:
Closes #827 
Closes #536 

Integration tests were attempted here, but hampered by our lack of ability to add SirTrevor widgets in our JS tests.

canvasindex


app/helpers/application_helper.rb
```diff
@@ -28,13 +28,37 @@ def iiif_drag_n_drop(manifest, width: '40px')
...
+
+  ##
+  # This method sends the message of which "canvas index" (zero-based) should be
+  # selected for a oembed viewer. For this it assumes that a
+  # SirTrevorRails::Blocks::SolrDocumentsEmbedBlock's first item has
+  # `iiif_canvas_id` and that `@id` conforms to the common DLSS `@id` format
+  # e.g. https://purl.stanford.edu/ab123cd4567/iiif/canvas/ab123cd4567_1
+  # @param [SirTrevorRails::Blocks::SolrDocumentsEmbedBlock] block
+  # @param [Integer]
+  def choose_canvas_index(sir_trevor_block)
```
cbeer: Ought we push this logic into `SirTrevorRails::Blocks::SolrDocumentsEmbedBlock`?

mejackreed: Yeah I thought of that, but wasn't sure if it made sense. This method at the moment is kind of assuming our own canvas URI syntax. So it felt a little strange enforcing that upstream, but do you think a monkey patch in Exhibits would be better?

cbeer: Got it; nah, this is fine for now.

Gemfile
```diff
@@ -126,3 +126,5 @@ gem 'csl-styles'
 gem 'acts-as-taggable-on'
 gem 'mods_display'
 gem 'slack-ruby-client'
+gem 'blacklight-oembed', '~> 0.2'
+gem 'rack-test', '0.7.0'
```
drh-stanford: can you add a comment here on why we're pinning to a specific version? so we can unpin with the issue is resolved down the line...

mejackreed: Sure thing. I added it in the commit message, but probably should have put it here too. 

Added. Thanks for catching.

PR #975, opened 2017-12-04
Add the ability to notify slack when an exhibit's publish state changes

### LINKED ISSUE #488: Need service team notification when an exhibit is published
As a Spotlight service team member, I need to be notified right away when an exhibit is published, so I can keep better track of new exhibits and provide quick exhibit review and support to curators. I also need notification when an exhibit is unpublished so I can reach out to the exhibit creator as necessary.

It would also be very helpful to be notified of **who** published or unpublished an exhibit.

The Spotlight at Stanford service team has a mailing list:
spotlight-service@lists.stanford.edu, although this creates a JIRA ticket, so if the slack suggestion below was do-able, it might be the preferred option

jkeck:
Closes #488 

published

---

un-published


mejackreed:
gif

config/settings.yml
```diff
@@ -33,3 +34,6 @@ harvestdor:
 gdor:
   service_url: https://dorfetcher-url
   skip_heartbeat: true
+slack_notifications:
+  default_channel: '#spotlight-service-team'
+  api_token: 'your-api-token-here'
```
camillevilla: 🔑? Does this `api-token` need to point at at environment variable @jkeck?

jkeck: @camillevilla 🤔 no, I don't think so.  This can be set in a `development.local.yml` and I'll submit a PR to shared_configs to add our exhibits-bot API key if this is accepted.

jkeck: sul-dlss/shared_configs#429 & sul-dlss/shared_configs#430 👀

app/models/concerns/exhibit_extension.rb
```diff
@@ -17,4 +19,11 @@ def required_viewer
     return viewer if viewer.present?
     Viewer.create(exhibit: self)
   end
+
+  private
+
+  def send_publish_state_change_notification
+    return unless saved_changes.keys.include?('published')
+    SendPublishStateChangeNotificationJob.perform_later(exhibit: self, published: published)
```
drh-stanford: This `published` variable is a little confusing to me. `Spotlight::Exhibit` has it as an ActiveRecord::Relation (i.e., `scope :published, -> { where(published: true) }`). In the Job code, you're just using it as a boolean flag. Maybe use `published.present?` here to indicate that it's just a flag? Or am I misunderstanding this code?

jkeck: `#published` on an instance of an Exhibit is a boolean while `#published` on the Exhibit class is a scope (I do understand that the scope and instance method being named the same thing is a bit confusing).

```ruby
Spotlight::Exhibit.published => ActiveRecord::Relation
Spotlight::Exhibit.new.published => Boolean
```

I'm not 100% sure saying effectively saying `true.present?` vs `false.present?` makes things any clearer here?

drh-stanford: ok that makes sense now. thanks.

PR #983, opened 2017-12-05
Add non-admin UI elements to i18n keys

camillevilla:
This PR moves strings in the user-facing UI, created during the Parker work cycle, with i18n keys. 

- Removes the bibliography service keys from early Exhibits / Spotlight 2016 work cycle
- Adds keys for the following features:
  - bibliography resource
  - Metadata modal download button
  - ~~search fields (Title, Subject, etc.)~~ --> split off to #1056

### Todo:
- [x]  update to Spotlight 1.2.0 / Blacklight 6.13 (#1012)
- [x] update keys
- [x] update CHANGELOG
- [x] major rebase 😭
~~- [ ] get search field labels working~~

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,8 +8,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ### Added
 ### Changed
+- Adds i18n keys bibliography resources and metadata display modal  #983
 ### Deprecated
 ### Removed
+- Removes i18n keys for deprecated bibliography service #983
 ### Fixed
 ### Security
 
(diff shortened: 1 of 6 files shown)


app/controllers/catalog_controller.rb
```diff
@@ -218,15 +218,15 @@ class CatalogController < ApplicationController
     # since we aren't specifying it otherwise.
 
     config.add_search_field('search') do |field|
-      field.label = 'Everything'
+      field.label = I18n.t('search_fields.keyword')
```
cbeer: We backported the I18n for search + sort fields into Blacklight 6.13 here: https://github.com/projectblacklight/blacklight/pull/1815

jkeck: So I believe that means the path forward on this is to upgrade to Blacklight 6.13.0 and then add keys under the Blacklight 18n namespace:

Issue #444, opened 2017-02-14
Provide support for multilingual text display


blalbrit:
For instance: support switching between Italian text and English text - both within the curator-supplied text of a page and the metadata labels.

blalbrit:
- [ ] add toggle to switch between languages (language selector)
- [ ] enable curators to provide translations for narrative text
- [ ] add i18n file for UI translations
- [ ] https://github.com/sul-dlss/exhibits/issues/952 (tag translation ticket)

ggeisler:
We decided that some of this work should be done in the Spotlight repo and so we've made new tickets there that cover much of what @blalbrit outlined in his comment above. So I'm closing this ticket to prevent potential confusion. I've also closed the Exhibits subtickets mentioned above that now have equivalent tickets in Spotlight. 

See https://github.com/projectblacklight/spotlight/issues/1912 for the story ticket in Spotlight, equivalent this this one, that lists the subtickets related to the feature. 

Of the tickets @blalbrit listed above, one does remain open in Exhibits (more appropriate here, I think): https://github.com/sul-dlss/exhibits/issues/952

anarchivist:
I accidentally deleted the comment that articulated all multilingual tickets linked from here - but I believe they should all be represented with the [`i18n`](https://github.com/sul-dlss/exhibits/labels/i18n) label.

PR #997, opened 2017-12-07
[WIP] Add locale-based routing and a locale picker in the navbar

### LINKED ISSUE #984: Add toggle to switch between languages (language selector)
As a user: 

* I should be able to see a control that allows me to select the language in which I wish to view an exhibit.
* Selecting a language should show me as much page content as possible translated for that language.

_Note:_ Translations will almost definitely need to be pulled from multiple sources (locales and db-backed content). Design can probably proceed on this, but I'm not sure when we should schedule the implementation yet.

cbeer:
This adds a slightly more sophisticated version of the locale picker from the blacklight demo.

Note that the locale picker only shows up if there are multiple options to pick; right now I've only added `en`, so this is effectively a no-op for our users.

Fixes #984 

screen shot 2017-12-07 at 08 08 30

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -22,4 +22,14 @@ class ApplicationController < ActionController::Base
     @feature_flags ||= FeatureFlags.for(current_exhibit)
   end
   helper_method :feature_flags
+
+  before_action :set_locale
+
+  def set_locale
+    I18n.locale = params[:locale] || I18n.default_locale
+  end
+
+  def default_url_options
+    { locale: I18n.locale }
+  end
 end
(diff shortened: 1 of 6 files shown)


jkeck:
We should validate that projectblacklight/blacklight#1830 isn't an issue for us.

ggeisler:
Based on our plan to implement #444 and some of its component tickets in Spotlight, I wonder if this work should be in Spotlight? Or is it something that we'll have to do in Spotlight in addition to Exhibits? 

One way other another, I think we'll need an equivalent locale switcher in Spotlight, unless it makes sense to offer the means to generate translations in Spotlight but then assume that Spotlight implementors need to do their own version of this PR to make the translations accessible to exhibit visitors.

PR #1004, opened 2017-12-08
reverse the order of the collection and collection druid for sorting …

### LINKED ISSUE #810: A-Z sort in collections facet not working correctly
In the following exhibit:
https://exhibits.stanford.edu/data

The A-Z sort is not working properly in the collection facet:
https://exhibits.stanford.edu/data/catalog?q=&search_field=searchUUIDa791febd-f7aa-43e1-97ab-26519cec01d3

mejackreed:
…sake

Closes #810 

This also would require a reindex of everything in production. I'm not sure of a better way to do this. Adding a duplicate field (instead of reversing) feels suboptimal too as users will now have two collection fields in there exhibit. We should just do the thing that we want at the end of the day.

## Correct 0-9 A-Z order by title (not druid)
screen shot 2017-12-08 at 1 46 20 pm

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -5,7 +5,7 @@ module ApplicationHelper
   # Collection titles are indexed as a compound druid + title; we need to
   # unmangle it for display.
   def collection_title(value, separator: '-|-')
-    value.split(separator).last
+    value.split(separator).first
   end
 
   def document_collection_title(value:, **)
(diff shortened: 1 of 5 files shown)


jkeck:
I'm wondering if we even use the druid in the exhibits case?  This may be a holdover from when we were using the SearchWorks indexing code to do this, and may not be necessary since we own our indexing code now.  I can look more into this when I have a bit of time (unless you get to it).

jkeck:
Yeah, it looks like the druid piece of this field is never used.  I don't want to hold this up, so I'll create a tech debt ticket about dealing with this later.

PR #1022, opened 2017-12-14
Show Mirador full screen button at smaller widths

### LINKED ISSUE #1017: Inconsistent full-screen icon in Mirador
The full-screen icon/capability works for some objects but not for others - see examples below:

Works: https://exhibits.stanford.edu/parker-2/catalog/pm049bv8195
screenshot 2017-12-13 12 17 38

Doesn't work/appear: https://exhibits.stanford.edu/parker-2/catalog/kz540zc7240
screenshot 2017-12-13 12 20 20

Could this have something to do with the width of the page-picker widget at time of instantiation? Weird if so, but that's the only difference I see between these two objects.

camillevilla:
Fixes #1017

This fixes a bug where the Mirador fullscreen button was hidden by the page picker widget. 

**Note:** @blalbrit this is quick fix that only applies to a Mirador viewer with 1 slot. Getting the CSS to work with multiple slots would be a more involved solution. 

## Fixtures
pm049bv8195
kz540zc7240

**Note:** annotation indexing must be enabled in settings.yml to get these longer canvas labels (e.g. "Front outside bindings") from the custom manifests

### Before

before_narrow
before_wide

### After
after_narrow
after_wide

### DIFF
--- a/app/assets/stylesheets/mirador_bundle.scss
+++ b/app/assets/stylesheets/mirador_bundle.scss
@@ -5,5 +5,16 @@
 .mirador-viewer {
   .page-select {
     margin-right: 5px;
+    max-width: 50%;
+  }
+
+  .window-manifest-navigation {
+    @media screen and (max-width: 768px) {
+      min-width: 42%;
+    }
+
+    @media screen and (max-width: 575px) {
+      min-width: 50%;
+    }
   }
 }


blalbrit:
Perfect @camillevilla - many thanks, and understood re: no support for multi-slot configurations.

B.

aeschylus:
Looks good to me. Merging.

PR #1031, opened 2017-12-19
Update changelog for 1.13.0 release.

jkeck:


jkeck:
Ugh, just realized we're missing a commit that we need to have before this release, will re-submit this later.

PR #1032, opened 2017-12-19
Only index collection title in the collection_with_title field.

### LINKED ISSUE #1020: Refactor collection_with_title field to not include the druid
We don't make use of the druid part of this field and have to provide helper methods handle the field parsing.  Since we own our own indexing logic completely now, we don't need to have this compound field (which is a holdover from the SearchWorks indexing pipeline).

### LINKED ISSUE #1029: Collection facet not showing collection name
I imagine this has to do w/ #1004

We'll need to address this (possibly by doing #1020), or revert that PR before our next release.

On exhibits-stage:
colection-facet-bug

jkeck:
Closes #1029 
Closes #1020

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Fixed
 - Fixed timeout issue on Add Items page for exhibits with large number of items #1013
 - Fixes a bug where newly created admins and curators were not receiving invitations #1012
+- Fixes an issue where the collection facet was sorting by druid and not by collection title #1004
 ### Security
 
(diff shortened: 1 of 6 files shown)


cbeer:
I'm not sure I understand how this fixes our deployment sequencing problem, unless you plan to also revert #1004 too?

jkeck:
@cbeer good catch, sorry, I'm still a little out of it from being sick.

I updated the helper w/ my intention as well as a comment about the behavior.  Let me know if that helps clear things up.

cbeer:
I've also thought of another reason we should probably just choose a new field name -- if the field is used to construct browse categories etc, the new indexing will break them.

jkeck:
Ah, good point. 👍 I'll close this, revert the previous PR, and re-open the issue.

PR #1040, opened 2017-12-20
Add help text to 'Add a new user' form

### LINKED ISSUE #998: Should we update the help text on the Add User admin screen
We have seen a few SUNet aliases as well as non-stanford addresses added.  We may want to add some help text that lets the users know the constraints of the kinds of users they can add to an exhibit.

camillevilla:
Closes #998 
WIP:
- [x] Make upstream change in Spotlight (waiting for projectblacklight/spotlight#1874)
- [x] Spotlight release: 1.2.0 --> 1.3.0?
- [x] Bump exhibits Spotlight version
- [x] Update this PR with Stanford-specific i18n key; remove JS

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ### Added
 - Added support for highlighting matching query terms from full-text content in search results #1030
+- Adds Stanford-specific helptext to the add admin/curator form #1040
 ### Changed
 ### Deprecated
 ### Removed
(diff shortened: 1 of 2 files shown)


jkeck:
Would it be possible for us to modify the Spotlight view code to reference an `i18n` key that we can override in our application (that may or may not require that we have some generic text in Spotlight, not sure off the top of my head how it behaves when not providing any text upstream)?

One of the benefits that this will provide is that it will make the internationalization of that content much easier if/when we need to consider internationalizing the admin interface.

camillevilla:
@jkeck that makes sense. I'll get started on something upstream.

PR #1042, opened 2017-12-21
Allow for exhibit specific manifest URL patterns to be used for our oembed tag.

### LINKED ISSUE #1038: Expose content_search implementation through to VT test exhibit
Once we have a version of the app up, we need to surface the content_search implementation through to the VT test exhibit so we can begin to experiment with it.

Currently, this is deployed within UAT; not sure if we need a feature flag to allow us to pull from `embed-uat` instead of `embed-prod`.

jkeck:
This allows us to configure, on a per exhibit basis, to use a different oEmbed environment.

Attempt to close #1038 

Hoping to start a conversation about the approach here.

In order to test this out, ensure your local exhibit's viewer is set to SUL Embed (the default) and set a IIIF url in the other tab to `https://sul-purl-uat.stanford.edu/{id}`. (note, if we want to use the Mirador viewer, we simply modify that URL to be `https://sul-purl-uat.stanford.edu/{id}/iiif/manifest`).

Then index an object that advertises our content search implementation in uat (e.g. cc842mn9348)

viewer-w-search-service

One side-note about how this works is that due to the current logic in `ManifestConcern`, this will still bail out and use the production Embed for objects that are not image-like (image, manuscript, map, book).  That's probably okay for our current purposes (given that this is intended to be temporary to allow us to begin poking at content search), but we may want to keep it in mind moving forward.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Added
 - Added support for highlighting matching query terms from full-text content in search results #1030
 - Adds Stanford-specific helptext to the add admin/curator form #1040
+- Added support for using the exhibit specific manifest URL configuration to configure the SUL Embed environment to be used #1042
 ### Changed
 ### Deprecated
 ### Removed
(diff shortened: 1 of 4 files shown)


camillevilla:
@cbeer do you have any additional comments?

PR #1043, opened 2017-12-23
Add a POC for indexing ALTO 2/3 OCR XML.

### LINKED ISSUE #1039: Full text indexing from OCR for discovery/hit highlighting
As an exhibit creator:

* When I add an SDR object with OCR files associated with it, I want the text within those OCR files to be searchable within the exhibit.
* OCR index should support (in priority order):
     * ALTO 3
     * ALTO 2
     * plain text

As a user:

* I want to be able to search the full text of a item, and find items that whose full text matches my search terms.

jkeck:
Closes #1039 

#### TODO:
- [x] Ensure XML we're processing has an ALTO namespace
- [x] Determine if checking the content-type of the response is acceptable (or should we refactor a bit to be able to use the mime-type described in the contentMetadata)
- [x] ~~Validate we're handling restricted content~~ (We're not going to worry about this at the moment.  We'll come up w/ a longer term strategy in sul-dlss/content_search#29)
- [x] Handle spaces/characters in file paths.

---

**Example:** _cc842mn9348_
cc842mn9348-highlight

### DIFF
--- a/app/models/full_text_parser.rb
+++ b/app/models/full_text_parser.rb
@@ -1,0 +1,61 @@
+# frozen_string_literal: true
+
+##
+# FullTextParser takes in a purl object resources and knows
+# how to find full-text resources (either plain-text or ALTO xml)
+# and can return an array of OCR strings (one element in the array for every OCR file).
+class FullTextParser
+  attr_reader :druid, :purl_object
+  delegate :public_xml, to: :purl_object
+
+  def initialize(purl_object)
+    @purl_object = purl_object
+    @druid = purl_object.bare_druid
+  end
+
+  def ocr_files
+    @ocr_files ||= xpath.flat_map { |xp| public_xml.xpath(xp) }
+  end
+
+  # TODO: Support .rtf (and .xml?) files, scrubbing/converting as necessary to get plain text
+  def to_text
+    ocr_files.map do |resource|
+      url = full_text_url(resource['id'])
+      content = full_text_content(url)
+      if ['application/xml', 'application/alto+xml'].include?(resource['mimetype'])
+        alto_xml_string_content(content)
+      else # plain text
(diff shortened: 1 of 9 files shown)


app/models/full_text_parser.rb
```diff
@@ -0,0 +1,61 @@
...
+    alto_ns = alto.namespaces.values.first { |ns| ns =~ %r{standards/alto/ns} }
+    namespace = { alto: alto_ns || 'http://www.loc.gov/standards/alto/ns-v3#' }
+    alto.xpath('//alto:String/@CONTENT', namespace).map(&:text).join(' ')
+  end
+
+  # full text in druid.txt named for druid (feigenbaum) and ALTO OCR xml in page resources
+  def xpath
+    [
+      "//contentMetadata/resource/file[@id=\"#{druid}.txt\"]",
+      "//contentMetadata/resource[@type='page']/file[@mimetype='application/xml' or @mimetype='application/alto+xml']"
```
jkeck: I could be easily convinced to pull these mime-types into a constant or a configuration.

spec/features/indexing_integration_spec.rb
```diff
@@ -137,6 +137,30 @@
...
+        'https://stacks.stanford.edu/file/cc842mn9348/EastTimor_CE-SPSC_Final_Decisions_2001_04b-2001_Sabino_Gouveia_Leite_Judgment_0001.xml'
+      ).to_return(body: fixture_file, status: 200)
+    end
+
+    it 'indexes all strings from the ALTO files into the full text field' do
+      fulltext = document[:full_text_tesimv]
+      expect(fulltext).to be_a Array
+      expect(fulltext.length).to eq 1
+      expect(fulltext.first).to start_with 'INTRODUCTION 1 The trial of'
+      expect(fulltext.first).to end_with 'rendering of the decision.'
```
jkeck: This test is now basically duplicated here and in the test for the class.  Not sure this one is necessary anymore, but kinda nice to know it works?  `/shrug`

PR #1045, opened 2018-01-10
Configure the full-text field to be disabled by default…

jkeck:
… (and only allow enabling on the list view).

### DIFF
--- a/config/environments/test.rb
+++ b/config/environments/test.rb
@@ -22,7 +22,7 @@ Exhibits::Application.configure do
 
   # Raise exceptions instead of rendering exception templates.
   config.action_dispatch.show_exceptions = false
-  
+
   # Disable request forgery protection in test environment.
   config.action_controller.allow_forgery_protection = false
   config.action_mailer.perform_caching = false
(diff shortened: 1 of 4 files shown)


jkeck:
@caaster the way this is implemented it is disabled for all views by default and can ONLY be enabled for the `List` view (standard search result view). This is akin to how the `Inline Map` field works except the only available view is `List` instead of `Item Details` view.

config/environments/test.rb
```diff
@@ -36,5 +36,6 @@
   config.active_support.deprecation = :stderr
 
   # Raises error for missing translations
-  config.action_view.raise_on_missing_translations = true
+  # Temporarily setting to false because spotlight has some missing keys that need to be fixed
+  config.action_view.raise_on_missing_translations = false
```
camillevilla: @jkeck is this ticketed anywhere?

jkeck: Ah, no, I was going to submit a PR (I have a local branch fixing some of them) but I bailed to get this out the door.  I'll add that ticket now. 👍

PR #1055, opened 2018-01-19
Add a feature flag for using the the PURL UAT environment for embed.

### LINKED ISSUE #1038: Expose content_search implementation through to VT test exhibit

jkeck:
Trying to close #1038 again (but using a feature flag this time).

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -45,7 +45,7 @@ module ApplicationHelper
   # @param [SolrDocument] document
   # @param [Integer] canvas_index
   def custom_render_oembed_tag_async(document, canvas_index)
-    url = document.first(blacklight_config.show.oembed_field)
+    url = context_specific_oembed_url(document)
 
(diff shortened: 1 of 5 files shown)


app/helpers/application_helper.rb
```diff
@@ -61,4 +61,14 @@ def custom_render_oembed_tag_async(document, canvas_index)
   def choose_canvas_index(sir_trevor_block)
     [sir_trevor_block.try(:items).try(:first).try(:[], 'iiif_canvas_id').try(:[], /\d*$/).to_i - 1, 0].max
   end
+
+  private
+
+  def context_specific_oembed_url(document)
```
jkeck: I'm having a hard time naming this.

PR #1060, opened 2018-01-25
Add support for full text indexing of Portuguese and Indonesian

### LINKED ISSUE #979: Support multilingual text in the same field
Virtual Tribunals has documents that may have running text in multiple languages (e.g. mostly Portuguese, with a few paragraphs of English interspersed). Investigate methods for enabling indexing and search for this kind of content

jkeck:
(in addition to the existing English support)

Closes #979 

## Document cb774wc3448
cb774wc3448-homens

### TODO
- [x] Resolve open questions listed in this PR.
- [x] Write code to handle rendering the highlight from the correct field.
- [x] Figure out testing approach
- [x] Limit to 5 highlight sections returned
- [x] Update sul-solr configs

### Open Questions
- ~~We currently intend to limit to 5 highlights.  However; since we're not returning fields for multiple languages, this 5 highlight limit applies per field (which means we can get 5 per language + 1).  Should we limit the number of highlights we'll return on the app side to handle this?~~
- Do we need to coordinate this w/ a re-index (of at least Feigenbaum content)?
- `SnowballPorterFilterFactory`: Should we just roll with this difference between the systems and see what happens?


solr/config/schema.xml
```diff
@@ -374,6 +380,55 @@
       </analyzer>
     </fieldtype>
 
+    <!-- Analyzed Text, English -->
+    <fieldtype name="text_en" class="solr.TextField" positionIncrementGap="10000" autoGeneratePhraseQueries="true">
```
jkeck: We're using a `positionIncrementGap` of 100 in content search, should we be doing the same here?

solr/config/schema.xml
```diff
@@ -374,6 +380,55 @@
...
+        <filter class="solr.RemoveDuplicatesTokenFilterFactory" />
+      </analyzer>
+    </fieldtype>
+
+    <!-- Analyzed Text, Portuguese -->
+    <fieldtype name="text_pt" class="solr.TextField" positionIncrementGap="10000" autoGeneratePhraseQueries="true">
+      <analyzer>
+        <tokenizer class="solr.WhitespaceTokenizerFactory" />
+        <filter class="solr.ICUFoldingFilterFactory"/>  <!-- NFKC, case folding, diacritics removed -->
+        <filter class="solr.SynonymFilterFactory" synonyms="synonyms.txt" ignoreCase="true" expand="false"/>
```
jkeck: I'm guessing we actually don't want to us the `SynonymFilterFactory` in these non-english fields?  (or will we need/want to add non-english synonyms later)

If we do need to remove this from the non-english fields, is it okay that it exists in the generic `text` field? (this would differ from how content search is setup)

solr/config/schema.xml
```diff
@@ -374,6 +380,55 @@
       </analyzer>
     </fieldtype>
 
+    <!-- Analyzed Text, English -->
+    <fieldtype name="text_en" class="solr.TextField" positionIncrementGap="10000" autoGeneratePhraseQueries="true">
+      <analyzer>
+        <tokenizer class="solr.WhitespaceTokenizerFactory" />
+        <filter class="solr.ICUFoldingFilterFactory"/>  <!-- NFKC, case folding, diacritics removed -->
+        <filter class="solr.SynonymFilterFactory" synonyms="synonyms.txt" ignoreCase="true" expand="false"/>
+        <filter class="solr.WordDelimiterFilterFactory"
```
jkeck: I noticed that we're using `WordDelimiterGraphFilterFactory` in content search.  `WordDelimiterFilterFactory` (used here) is deprecated.  Not sure if this will cause any issues currently, but was wondering if we should update exhibits to use the non-deprecated version.

solr/config/schema.xml
```diff
@@ -248,6 +248,9 @@
 
     <!-- exhibits fields -->
     <field name="full_text_search" type="text" indexed="true" stored="true" termVectors="true" termPositions="true" termOffsets="true" multiValued="true" />
+    <field name="full_text_search_en" type="text_en" indexed="true" stored="true" termVectors="true" termPositions="true" termOffsets="true" multiValued="true" />
+    <field name="full_text_search_pt" type="text_pt" indexed="true" stored="true" termVectors="true" termPositions="true" termOffsets="true" multiValued="true" />
+    <field name="full_text_search_id" type="text_id" indexed="true" stored="true" termVectors="true" termPositions="true" termOffsets="true" multiValued="true" />
     <field name="full_text_unstem_search" type="textNoStem" indexed="true" stored="true" multiValued="true" />
```
jkeck: Do we need to create `full_text_unstem_search` fields for each language as well?

solr/config/schema.xml
```diff
@@ -248,6 +248,9 @@
 
     <!-- exhibits fields -->
     <field name="full_text_search" type="text" indexed="true" stored="true" termVectors="true" termPositions="true" termOffsets="true" multiValued="true" />
+    <field name="full_text_search_en" type="text_en" indexed="true" stored="true" termVectors="true" termPositions="true" termOffsets="true" multiValued="true" />
```
jkeck: Not totally sure about the naming of these fields.  It seems like the right approach to make the various `full_text_search_*` fields for each language and then copy from full_text_search into them.  Totally willing to entertain other options.

solr/config/schema.xml
cbeer: No, I think that should be unnecessary. I think we only need language-specific fields when we're doing stemming, lemmatization, or language-specific tokenization.

solr/config/schema.xml
cbeer: 👍 it'll be easier now than later.

solr/config/schema.xml
cbeer: I think the synonyms we've defined so far are just stolen from searchworks and not language-specific (and unlikely to be interesting in the near future, either.)

solr/config/schema.xml
```diff
@@ -361,16 +367,85 @@
 
     <!-- Analyzed Text, general case -->
     <fieldtype name="text" class="solr.TextField" positionIncrementGap="10000" autoGeneratePhraseQueries="true">
-      <analyzer>
+      <analyzer type="index">
+        <tokenizer class="solr.WhitespaceTokenizerFactory" />
+        <filter class="solr.ICUFoldingFilterFactory"/>  <!-- NFKC, case folding, diacritics removed -->
+        <filter class="solr.SynonymFilterFactory" synonyms="synonyms.txt" ignoreCase="true" expand="false"/>
+        <filter class="solr.SnowballPorterFilterFactory"/>
```
jkeck: One of the big differences between our configurations and the ContentSearch configurations is we were already using the `SnowballPorterFilterFactory` for our full-text fields.  Removing this in our configurations seems to make the existing full-text search (used be Feigenbaum) to return vastly different highlighting (and potentially search results).

jkeck:
I'm not sure if we'll NEED coordinate this w/ a re-index of any full-text content in order for this to work (@cbeer perhaps you can shed some light).  I've done a bit of poking around our index to see what the extent of the exhibits/documents that have existing full-text content.

- Feigenbaum is the only exhibit that has full text content indexed (not surprising).
- The primary concern will be the affect on Feigenbaum browse categories that are constructed w/ query terms.
  - 14/22 Feigenbaum browse categories are constructed w/ a query term.
- **Current Feigenbaum browse categories and counts**
  - stanford-university-medical-experimental-computer-sumex: 208
  - book-computers-and-thought: 74
  - epam: 258
  - dendral: 1474
  - expert-systems-general: 327
  - expert-systems-by-heuristic-programming-project: 118
  - air-force: 54
  - book-fifth-generation: 701
  - stanford-heuristic-programming-project: 78
  - the-mycin-experiments: 35
  - pulmonary-function-advisory-system-puff: 7
  - book-handbook-of-ai: 220
  - heuristic-adaptive-surveillance-process-hasp: 69
  - computer-science-department: 56
  - darpa: 577
  - book-rise-of-the-expert-company: 23
  - herbert-simon: 162
  - handwritten-notes: 101
  - sound-recordings: 119
  - videos: 66
  - all-exhibit-items: 16417
  - expert-systems: 156

ggeisler:
> We currently intend to limit to 5 highlights. However; since we're not returning fields for multiple languages, this 5 highlight limit applies per field (which means we can get 5 per language + 1). Should we limit the number of highlights we'll return on the app side to handle this?

@jkeck: I'd suggest we always limit to 5 highlights (for space and consistency reasons), no matter how many language fields there are. The downside might be that all 5 will be in one language, even though there are full-text results in more than one language, but that seems okay to me, since the highlights are only intended to be a preview and the user presumably understands that there might be more that are not visible.

jkeck:
All the query based (e.g. not purely facet based) browse categories had the same number of documents before and after this update on stage (and a re-index).  I think we should be 👌 to roll this out to production w/o impacting the current full-text indexing landscape.

aeschylus:
Adding a link to: https://github.com/sul-dlss/sul-solr-configs/pull/118

PR #1073, opened 2018-01-30
Handle the case where no solr response is available in context…

### LINKED ISSUE #1072: Use of search widget currently breaks page display
@jkeck says: I _think_ that full text previews just may not work when pulling search results into feature pages (but I’m guessing that’s probably fine). And if that’s the case, the fix is two characters.

jkeck:
…(e.g. SirTrevor search results widget).

Fixes #1072


jkeck:
Note that full-text highlighting will not work in the search results widget.  If we want that behavior to be possible, we should ticket it separately (and may be technically pretty difficult)

PR #1074, opened 2018-01-30
Make the oembed viewer the full-width of the item show page; fixes #1062

### LINKED ISSUE #1062: Make viewer full-width of item show page
This has been discussed before and I can't remember if there is a reason we haven't done it, but unless there are objections I suggest we update the layout of the item show page to give the viewer full-width. This will match the layout of the item show page in an exhibit where the Mirador viewer is configured as the viewer, and it will provide more usable space in the viewer for zooming, show the info panel, etc.

Here's a rough idea of how it will look with the change:

----
viewer-full-width

----
Getting to that includes these updates:

* Make the viewer take 100% of the width, from 50%.
* Increase the height of the viewer (the current height is arguably too limited and that limitation will be more noticeable with the increased width). It's hard for me to suggest the optimal height without actually implementing it, so I suggest we experiment with the height setting, but as a start I'd suggest `600px` (or possibly using `65vh` for a more responsive height).
* Change the width of the metadata `dl` so it doesn't span the full-width of wide pages. I'd suggest using `width: 85ch;` (it looks like it is set to `50%` on the page when using the Mirador viewer, but I think `width: 85ch;` might work better at most viewport widths).
* Add spacing between the bottom of the viewer and the start of the `dl` (20px), but that might be better addressed in #1063, which suggests adding a Purl link below the viewer, if we implement that ticket.

### LINKED ISSUE #1063: Add Purl link below viewer on item show page
Regardless of what we decide about the More Info panel in the viewer, the Purl link for an item seems important enough that a user should be able to find it more easily than opening a panel in the viewer. The mockup below suggests placing it below the viewer, aligned to the left-edge of the viewer. This placement would work for both the UV viewer and the Mirador viewer (which has the IIIF icon below it but aligned to the right-edge).

(The mockup below assumes we are also implementing #1062 but should be feasible even if we decide not to implement that ticket, although the bottom margin would probably not be necessary.)

----
purl-link

----

Example CSS for styling:

```
.purl-link {
  font-size: 0.9em;
  margin-bottom: 2 * $padding-large-vertical;
}
```

cbeer:
Also, add a purl link below the viewer; fixes #1063.

 ## After
viewer full width

## Before
viewer_half_width

### DIFF
--- a/app/assets/stylesheets/modules/blacklight_overrides.scss
+++ b/app/assets/stylesheets/modules/blacklight_overrides.scss
@@ -22,6 +22,11 @@
   #sidebar {
     display: none;
   }
+
+  .dl-horizontal {
+    margin-top: 2 * $padding-large-vertical;
+    width: 85ch;
+  }
 }
 
 .thumbnail {
@@ -29,13 +34,6 @@
 }
 
 #document {
-  .oembed-widget {
-    @extend .col-md-6;
-    @extend .col-sm-12;
-
-    padding-left: 0;
-  }
-
   h1 {
     font-size: $font-size-h3;
   }
--- a/app/assets/stylesheets/modules/blacklight_overrides.scss
+++ b/app/assets/stylesheets/modules/blacklight_overrides.scss
@@ -24,6 +24,7 @@
   }
 
   .dl-horizontal {
+    float: none;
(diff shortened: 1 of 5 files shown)


camillevilla:
Looks good to me. Throwing a before screenshot in here for the record.

PR #1086, opened 2018-02-07
Explicitly cast the PurlFetcher config settings to a hash.

jkeck:
For whatever reason this seems to be required to work correctly on the server (e.g. wasn't necessary for me running the code locally).


app/models/dor_harvester.rb
```diff
@@ -96,7 +96,7 @@ def harvestdor_resource(druid)
 
   def harvestdor_indexer
     @harvestdor_indexer ||= Harvestdor::Indexer.new(
-      purl_fetcher: Settings.purl_fetcher,
+      purl_fetcher: Settings.purl_fetcher.to_h,
```
mejackreed: huh.. interesting.

PR #1087, opened 2018-02-09
Don't index empty date part values

cbeer:


lib/traject/dor_config.rb
```diff
@@ -97,7 +97,7 @@
     accumulator.concat(dates.map(&:values).flatten)
 
     part = imprint.send(:parts_element, value)
-    accumulator << part if part
+    accumulator << part unless part.blank?
```
jkeck: I'm not sure if rubocop is going to complain about this or not, but I think the preferred style is `if thing.present?` in this case.

cbeer: 👍

PR #1098, opened 2018-02-20
Make (Parker) thumbnail bgcolor same as theme body bgcolor

### LINKED ISSUE #1097: Update Parker theme to match thumbnail bgcolor with body color
It's only evident if an image doesn't take up all the space in its thumbnail container, but the default Spotlight thumbnail background color is white, which doesn't look ideal with Parker's off-white body background color:

item_row_widget___parker_library_on_the_web_-_stanford_university_libraries

ggeisler:
Fixes #1097 

Before:

item_row_widget___parker_library_on_the_web_-_stanford_university_libraries

After:

after

### DIFF
--- a/app/assets/stylesheets/modules/parker.scss
+++ b/app/assets/stylesheets/modules/parker.scss
@@ -1,4 +1,5 @@
-body {
+body,
+.thumbnail {
   background-color: $color-beige-05;
 }
 


ggeisler:
No, I don't think so. I thought about doing it in a more generic way where the thumbnail always gets the same background color as the body, which would probably be fine most of the time, but I could see a scenario where someone might want them to be different. So this approach seems better for not making someone have to override that assumption.

PR #1104, opened 2018-02-23
Only index xml OCR w/ a role of "transcription".

jkeck:


app/models/full_text_parser.rb
```diff
@@ -55,7 +55,8 @@ def alto_xml_string_content(content)
   def xpath
     [
       "//contentMetadata/resource/file[@id=\"#{druid}.txt\"]",
-      "//contentMetadata/resource[@type='page']/file[@mimetype='application/xml' or @mimetype='application/alto+xml']"
+      "//contentMetadata/resource[@type='page']/file[@role='transcription']\
+[@mimetype='application/xml' or @mimetype='application/alto+xml']"
```
jkeck: ¯\\_(ツ)_/¯ on this indentation

cbeer: I'd vote to just raise the line length.

jkeck: Hmm, I think it'll end up being pretty long, what about just ignoring it for this method?

mejackreed:
Let's wait just a bit on this until we can resolve the data in prod?

cc @cbeer

app/models/full_text_parser.rb
cbeer: 👍

cbeer:
Data updates should be in place.

PR #1116, opened 2018-03-08
increase column width for document thumbnails in the document header. Fixes #1114

### LINKED ISSUE #1114: In list view, title of object overlaps object thumbnail
This was reported by a user and it would be great if it could be fixed.
screen shot 2018-03-06 at 3 57 13 pm

aeschylus:
Closes #1114

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Deprecated
 ### Removed
 ### Fixed
+- Prevent list view titles from overlapping image thumbnails #1116
 ### Security
 
 ## [1.16.2] - 2018-03-07
--- a/app/assets/stylesheets/modules/blacklight_overrides.scss
+++ b/app/assets/stylesheets/modules/blacklight_overrides.scss
@@ -42,6 +42,7 @@
 
 .document-thumbnail img {
   max-height: 200px;
+  max-width: 100%;
 }
 
 #documents {
--- a/app/views/catalog/_exhibits_document_header_default.html.erb
+++ b/app/views/catalog/_exhibits_document_header_default.html.erb
@@ -2,10 +2,10 @@
   <div class='col-xs-3 col-md-2'>
     <%= render_document_partial(document, 'thumbnail', document_counter: document_counter) %>
   </div>
-  <div class='col-xs-8 col-md-9'>
+  <div class='col-xs-7 col-md-9'>
     <%= render_document_partial(document, 'index_header', document_counter: document_counter) %>
   </div>
-  <div class='col-xs-1'>
+  <div class='col-xs-2 col-md-1'>
     <% exhibit_specific_manifest = document.exhibit_specific_manifest(current_exhibit.required_viewer.custom_manifest_pattern) %>
 
     <%= iiif_drag_n_drop(exhibit_specific_manifest, width: '30') if exhibit_specific_manifest %>


ggeisler:
@jkeck Yeah, I looked at the updated code yesterday and it appeared to me to only affect the header (that holds the item title) so it seemed okay to me. I also applied the changes in web inspector and it looked fine, though it's hard to feel 100% confident without trying on results pages with varying sizes of thumbnails. 

So if @aeschylus feels confident about it it seems safe enough to merge and I can double check on published exhibits as soon as it's deployed to be sure (or we could check on -stage first if that's feasible).

ggeisler:
Though looking at this again, it doesn't seem to fix the problem at viewports less than `lg` (991px wide or less). So I guess we could either merge this PR to fix the immediate problem on desktop viewports and create a new ticket for a more long-term fix for complete responsiveness, or @aeschylus could revisit things to address all viewport widths.

The only reason I am not advocating more strongly to fix it at all viewports now is we have lots of places where our responsiveness is not good at mobile sizes, so I'm not sure whether that should be addressed here or as part of a larger focus on better mobile responsiveness.

aeschylus:
Here's a gif with the fix:
browse list header

aeschylus:
As Gary points out, there is still a small overlap at lower widths. I can play with it a bit more to resolve that. Should be a simple fix.

ggeisler:
@aeschylus Be sure to test with this page: https://exhibits.stanford.edu/tel-aviv/browse/black-white-postcards-pre-1948?view=list because it serves as a good example of thumbnails that are max-width for the thumbnail space and thus are more likely to overlap with the title text.

aeschylus:
@ggeisler I've reverted the column width values and forced the image to fit within its column. The problem was letting its height determine its width rather than the width of the grid column. Does this look alright? 
browse list header 2

It's good down to 375px, where the IIIF icon pops down under the heightless image. I can add some bottom padding to the image to prevent that, but then it will bump down the metadata and sit under the image on the left @<375.

aeschylus:
Here's an update with the column allowance expanded for the IIIF icon under 375. It doesn't squeeze out the rest of the content as much as I had feared. This seems like the best choice to me.

browse list header 2

ggeisler:
@aeschylus That looks good, but before we go with this I'm curious about something else that might be a factor here. Maybe I'm missing something, but I'm wondering if the original CSS is not the problem at all and it has something to do with the thumbnail images we're using. Here's what confuses me:

Some exhibits, such as Fitch and Parker, look fine; their thumbnails always fit and there is no overlap at different breakpoints (except maybe at the very narrowest viewport possible).

Other exhibits, such as Tel Aviv, Travel Through Time: Japan, and Images of Rome, have the overlap problem that instigated this ticket. But on those exhibits the thumbnail is noticeably larger. 

When I look closer with web inspector there seems to be a discrepancy in what size thumbnails those exhibits are bringing back:

Well-behaved example (Fitch) brings back a 100px square:

fitch

Problem example (Tel Aviv) brings back (it appears) a 400px square, even though (again, unless I'm missing something) we're requesting the thumbnail from Stacks in the same way:

tel-aviv

So, if we were getting 100px square thumbnail images consistently for all exhibits, I don't think we would see the overlap problem at all, even with our current CSS.

aeschylus:
That is true (that the thumbnail size is what puts things over the edge). Our current CSS tried to fix this by putting an arbitrary cap on the height, but depending on the aspect ratio of the image, a large enough image could still bleed over. Capping the height also doesn't play nicely at lower breakpoints, when the width of the column is less than the effective capped width of the image.

ggeisler:
My point is: why aren't we getting back 100px square thumbnails consistently?

I don't think we need to change the CSS at all if we can count on getting 100px thumbnails. That size is all we need for list view. We don't ever want large than that in list view.

cbeer:
@ggeisler that seems like a bug in our handling of no-download, perhaps. Let me get a policy clarification on that first.

jkeck:
So sul-dlss/stacks#340 seems to have addressed this.  Should we still make this column change or is it unnecessary now?

ggeisler:
I think we still want the `max-width: 100%;` change. It won't have a noticeable effect except at very small viewports, but is useful there.

I'm not sure we need any `col` changes, though. @aeschylus can comment on that part based on our conversation in Slack.

aeschylus:
This is what I see in production after fully clearing cache. The images behave as you would expect given 100px thumbs, but a slight overlap and whitespace issue remains:

screen shot 2018-03-09 at 10 31 52 am

Max-width will fix this, but the thumbs will certainly become small. I'll show the effects with and without the column change. I believe with max-width, it will still be necessary to add a column adjustment to prevent the IIIF icon from popping down, but let me test.

jkeck:
@aeschylus let us know if you get a chance to test this, so we can :shipit:.

aeschylus:
Hey @jkeck. I double checked this and the IIIF icon popping down is still happening at mobile portrait sizes without adjusting the columns. I recommend we merge this in its current form (with the columns adjusted).

camillevilla:
Thanks, @aeschylus. Looks great! Throwing some new screenshots in here for good measure:

## Chrome
smallest width in Chrome

## Firefox
^-- even smaller!
smallest width in in Firefox

aeschylus:
@camillevilla Thanks for the changelog update!

PR #1130, opened 2018-03-24
Always deploy from the master branch

cbeer:


jkeck:
I'm a bit on the fence about this.  We've typically tried to have a tagged release cycle for this that has related CHANGELOG updates that we can point PSM to.

What would folks thing about using something like:

```ruby
set :branch, proc { `git describe --tags --abbrev=0`.chomp }.call
```

which should give us the most recent tag in the given branch.

This would allow us to automate our deploys (to a certain extent) and still be intentional about having tagged release even for dependency updates (which we've been pretty good about so far).

mejackreed:
Are we able though to actually deploy to a tag? For some reason I thought that didn't work for us?

jkeck:
We weren't able to when we were using Squash.  Now that we're using Honeybadger it works fine.

Issue #1135, opened 2018-04-03
Embed widget in pages does not load properly for single image manifests


anarchivist:
For objects with single images, we are trying to instantiate the sul-embed viewer with `canvas_index=1`. Instantiating the UV is failing for single image items.

Examples:

* https://exhibits.stanford.edu/Welte-Mignon/feature/eugenie-adam-benard
* https://purl.stanford.edu/jg699wz4435

PR #1139, opened 2018-04-03
Pass the canvas id through to the embed viewer instead of trying to c…

cbeer:
…alculate a canvas index.

~Waiting for https://github.com/sul-dlss/sul-embed/pull/872.~

Fixes https://github.com/sul-dlss/exhibits/issues/1135.

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -51,15 +51,11 @@ module ApplicationHelper
   end
 
   ##
-  # This method sends the message of which "canvas index" (zero-based) should be
-  # selected for a oembed viewer. For this it assumes that a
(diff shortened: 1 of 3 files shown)


app/helpers/application_helper.rb
```diff
@@ -57,9 +57,9 @@ def custom_render_oembed_tag_async(document, canvas_index)
   # `iiif_canvas_id` and that `@id` conforms to the common DLSS `@id` format
   # e.g. https://purl.stanford.edu/ab123cd4567/iiif/canvas/ab123cd4567_1
   # @param [SirTrevorRails::Blocks::SolrDocumentsEmbedBlock] block
```
jkeck: I think the comment above is now invalid.

cbeer: :+1

PR #1149, opened 2018-04-17
update Spotlight and dependencies

mejackreed:


jkeck:
Wondering if we should add something to the CHANGLOG about the Spotlight update, since it's likely the only changes will be coming from the Spotlight update (and would be helpful for to describe what's changing in this release).

On a slightly unrelated note, I see that we never tagged a 1.7.3 release and it's not represented in our releases list.  I'll do that before we cut a new release w/ this update.

mejackreed:
@jkeck thanks for the feedback. I've updated the PR and added release notes. Let me know if I can do anything wrt the missing tag.

mejackreed:
@caaster gave the 👍

spec/controllers/application_controller_spec.rb
```diff
@@ -4,14 +4,17 @@
 
 describe ApplicationController do
   describe '#feature_flags' do
+    let(:controller) { described_class.new }
```
cbeer: I think this may be the same as the implicit subject (perhaps also already named controller?)

mejackreed: nice catch. Updated accordingly.

PR #1153, opened 2018-04-20
Adds configurable async javascript to improve time to first paint

mejackreed:
Since we include all of the JavaScript into a large bundle, lets load it
async so that we increase the page time to first paint for users.

https://developers.google.com/web/tools/lighthouse/audits/blocking-resources

## Before this pull request
screen shot 2018-04-20 at 4 36 18 pm

## Using async
screen shot 2018-04-20 at 4 36 27 pm


mejackreed:
Yep there are implications. Ideally we would like to do this everywhere, but because we set `config.assets.debug = true` in dev and test, this fails as we can't guarantee an order of assets

https://github.com/sul-dlss/exhibits/blob/master/config/environments/development.rb#L46

cbeer:
👍

PR #1164, opened 2018-05-01
Upgrade to Spotlight 2

mejackreed:
Remaining items:

 - [x] update `spotlight_initializer.rb` to use `Spotlight::UploadFieldConfig`


config/routes.rb
```diff
@@ -11,48 +11,60 @@
...
+    # this has to come before the Blacklight + Spotlight routes to avoid getting routed as
+    # a document request.
+    resources :exhibits, path: '/', only: [] do
+      get "catalog/range_limit" => "spotlight/catalog#range_limit"
+    end
 
-  mount Blacklight::Engine => '/'
-  resources :solr_documents, only: [:show], path: '/catalog', controller: 'catalog'
+    concern :searchable, Blacklight::Routes::Searchable.new
+    # mount Blacklight::Engine => '/'
```
cbeer: Remove these commented out lines?

cbeer:
For our first release, do we want to limit `i18n_locales` to just languages we know we care about or expect to have full translations for?

config/routes.rb
mejackreed: 👍 done.

mejackreed:
I could go either way on that. I'm kind of fine with people using the feature and finding new things with it. Do you see some other issue that might arise?

jkeck:
I'm in the mind that we roll w/ what Blacklight provides (which is what I believe we've done here)

PR #1166, opened 2018-05-03
Adds thumb/image indexing for virtual image objects

### LINKED ISSUE #484: Virtual objects don't display in list, gallery, masonry or slideshow view
When I am in list, gallery, masonry or slideshow view for any virtual object in an exhibit, there is no image displayed. 

NOTE 1: in slideshow view, I see a blank tile with the word "missing" in it, in place of the virtual object. When I view these same virtual objects in SearchWorks, I see an image displayed, regardless of the view selected. 

NOTE 2: Each virtual object DOES display images in Spotlight, at the item record view. 

Here are two virtual object druids, that are in an unpublished exhibit (https://exhibits.stanford.edu/reid-dennis) I use for testing/training -- (these druids are also available in SearchWorks):
ws947mh3822
fj384zr2364

I expect an image to be displayed for all virtual object druids in any Spotlight exhibit, regardless of the view selected.

mejackreed:
Fixes #484 

This commit updates the traject indexing code allowing for the logic
to process SDR "virtual objects" so that Exhibits can display
appropriate thumbs throughout.

## Before
screen shot 2018-05-03 at 10 47 25 am
## After
screen shot 2018-05-03 at 10 48 14 am


lib/traject/macros/dor.rb
```diff
@@ -16,11 +16,18 @@ def resource_images_iiif_urls
...
         next if content_metadata.nil?
 
-        images = content_metadata.xpath('resource/file[@mimetype="image/jp2"]')
-        jp2s = images.select { |node| node.attr('id') =~ /jp2$/ }
+        # Select conventional file images or virtual external ones
+        images = content_metadata.xpath(
+          '(resource/file[@mimetype="image/jp2"] | resource/externalFile[@mimetype="image/jp2"])'
+        )
+        # Allow for selection of conventional ids and fileId for virtual objects
+        jp2s = images.select { |node| (node.attr('id') || node.attr('fileId')) =~ /jp2$/ }
```
jkeck: Would it make sense to do a `collect` here so that `jp2s` is already the id that we want (and saves us from having to duplicate the logic below).  Just a thought.

jkeck: Ah, I see, we're also getting the `objectId` attribute below...  This feels a bit nasty, but don't have a succinct solution off the top of my head to suggest.

lib/traject/dor_config.rb
```diff
@@ -148,15 +148,16 @@
 each_record do |resource, context|
   content_metadata = resource.public_xml.at_xpath('/publicObject/contentMetadata')
   next if content_metadata.blank?
-  images = content_metadata.xpath('resource/file[@mimetype="image/jp2"]')
+  # Select conventional file images or virtual external ones
+  images = content_metadata.xpath('(resource/file[@mimetype="image/jp2"] | resource/externalFile[@mimetype="image/jp2"])')
   thumbnail_data = images.first { |node| node.attr('id') =~ /jp2$/ }
```
jkeck: Wouldn't this also have to check `fileId`?

PR #1167, opened 2018-05-03
update config initializer to Spotlight 2.0

mejackreed:
Not sure if I'm doing this right here.

Also have questions about the migration path for `upload_title_field`. I'm not sure that can be an array, no?


jkeck:
Happy to pair up w/ you on this.  Been reminding myself about this code a bit more.  We can probably stick w/ Arrays if we want (just to keep compatibility w/ what we had before).  The code that handles this field references all handle strings or arrays, so it looks like we'll be fine either way.

Another small addition is we may want to use `i18n` keys for our labels following the pattern that would have been generated `spotlight.search.fields.{field_name}` for us in the initializer.

PR #1177, opened 2018-05-11
Update Spotlight Upload Field configurations.

jkeck:
Currently, adding/editing titles/dates do not work for uploaded documents.

This approach adds a new field (following the naming convention for other upload fields) and uses solr's copyFields instead.

This approach will also require that we run a pre deploy data migration that populates this new field from existing data (for the 836 existing documents).

https://solr-host/solr/exhibits_prod/select?fq=spotlight_resource_type_ssim:spotlight/resources/uploads&wt=json&fl=id,title*,date*

### DIFF
--- a/config/initializers/spotlight_initializer.rb
+++ b/config/initializers/spotlight_initializer.rb
@@ -1,6 +1,6 @@
 Spotlight::Engine.config.upload_title_field = Spotlight::UploadFieldConfig.new(
-  field_name: %w(title title_full_display title_display title_245_search title_sort),
-  label: -> { I18n.t(:'spotlight.search.fields.title') }
+  field_name: :spotlight_upload_title_tesim,
+  label: -> { I18n.t(:'spotlight.search.fields.spotlight_upload_title_tesim') }
 )
 Spotlight::Engine.config.upload_fields = [
   Spotlight::UploadFieldConfig.new(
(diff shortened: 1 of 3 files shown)


jkeck:
Something along the lines of might work as a data migration

```ruby
Spotlight::Resources::Upload.find_each do |upload|
  data = upload.data
  upload.data = data.merge(
    'spotlight_upload_title_tesim': data['title']
    'spotlight_upload_date_tesim': data['date']    
  ).delete_if{ |_, v| v.nil? }

  upload.save_and_index
end
```

config/initializers/spotlight_initializer.rb
```diff
@@ -13,7 +13,7 @@
     label: -> { I18n.t(:'spotlight.search.fields.spotlight_upload_attribution_tesim') }
   ),
   Spotlight::UploadFieldConfig.new(
-    field_name: %w(date spotlight_upload_date_tesim date_sort),
```
mejackreed: Do we need to support the `date` field still?

config/initializers/spotlight_initializer.rb
```diff
@@ -1,6 +1,6 @@
 Spotlight::Engine.config.upload_title_field = Spotlight::UploadFieldConfig.new(
-  field_name: %w(title title_full_display title_display title_245_search title_sort),
```
mejackreed: Do we need to support the `title` field still?

config/initializers/spotlight_initializer.rb
jkeck: No, because its not a real solr field.

config/initializers/spotlight_initializer.rb
jkeck: No, because its not a real solr field.

camillevilla:
Validated this works in stage:

## Before 
upload_not_persisting
Changes to title etc. not persisting

## After
file renamed
Our file with updated title, attribution, and date

PR #1214, opened 2018-07-13
Update i18n key for Create Exhibit link

### LINKED ISSUE #1213: Update "Create new exhibit" link text
The link for creating a new exhibit in the user menu in the brandbar, when you are a superadmin, has (I'm pretty sure) changed and now just reads "Create". I checked in the latest Spotlight master and it still says "Create new exhibit" there, which is what I think it used to say in Exhibits. I don't really understand what's going on (e.g. why the locale key is missing in Exhibits if it is in Spotlight) but I do think we want the link label to read "Create new exhibit".

fullscreen_7_12_18__4_42_pm

Not a high priority because only superadmins see it and there is another way to get to the new exhibit form (sidebar link on the site landing page) but we should still fix.

jkeck:
Fixes #1213

### DIFF
--- a/app/views/_user_util_links.html.erb
+++ b/app/views/_user_util_links.html.erb
@@ -12,7 +12,7 @@
         <% end %>
         <% if can? :create, Spotlight::Exhibit %>
         <li class="divider"></li>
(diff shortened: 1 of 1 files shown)


jkeck:
@camillevilla not sure if this is the solution you were thinking.  Feel free to close this and submit a PR w/ your change.

PR #1240, opened 2018-09-07
Remove unused favicon

jcoyne:
The real favicon is in app/assets/images/favicon.ico


jkeck:
I think the intent is for the duplicate favicon to be served up under development so our local logs aren't polluted with 404s for the favicon.  Not sure if that holds true in modern rails. ¯\\_(ツ)_/¯

PR #1266, opened 2018-11-02
Adds a spec that instantiates Mirador

mejackreed:
This should have caught the issue we saw on 11/1/2018

`#ThisIsFine`

### DIFF
--- a/app/helpers/mirador_helper.rb
+++ b/app/helpers/mirador_helper.rb
@@ -16,6 +16,7 @@ module MiradorHelper
           "location": 'Stanford University'
         }
       ],
+      "language": 'en',
       "windowObjects": [{
         "loadedManifest": manifest,
         "canvasID": canvas,
--- a/spec/features/viewers_spec.rb
+++ b/spec/features/viewers_spec.rb
@@ -88,6 +88,11 @@ describe 'Viewers', type: :feature do
       expect(page).to have_css 'iframe[src*=mirador]'
       expect(page).not_to have_css '.oembed-widget'
     end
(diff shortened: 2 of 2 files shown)


app/helpers/mirador_helper.rb
```diff
@@ -16,6 +16,7 @@ def mirador_options(manifest, canvas, exhibit_slug)
           "location": 'Stanford University'
         }
       ],
+      "language": 'en',
```
jkeck: Should this be the `i18n.locale`?  I'm not sure how this plays in JS land (and if that'll end up causing us issues), but just wondering since actually use `i18n` with multiple languages in this app. Thanks for the clarification.

mejackreed: No I think we need to hardcode `en` here. If we don't, the Mirador i18n library will try and request whatever the browser `Accept-Language` is set to. This will cause problems, and potentially try and request `en-US` causing 404's.

mejackreed: Ok.. now that I think about this.. I guess this wouldn't matter for this case. But could cause some problems where `en` is not the language. Mirador doesn't have `it` or `pt-BR`.

aeschylus: Do we have a way of supplying those? I think the lookup is helpfully dumb in this case. It will look for those exact letters, so if we provided those files, it would pick them up. Not sure this is essential for adding this test though.

mejackreed: `/shrug`

PR #1267, opened 2018-11-05
Update dependencies

cbeer:


Gemfile.lock
```diff
@@ -415,7 +415,7 @@ GEM
       iso-639
       nokogiri
       nom-xml (~> 1.0)
-    mods_display (0.5.1)
+    mods_display (0.6.0)
```
mejackreed: This is fine, right?

cbeer: Yeah, it should be pretty small.. I think it's only a minor bump because of the new subtitle accessor?

PR #1296, opened 2018-12-03
Update upload_title_field solr configuration to remove a non-solr field

cbeer:
The conversation in #1177 makes me suspect we intended to remove the `title` field entirely, and it was mistakenly restored in #1258 as copy-pasta from the previous code. In any case, indexing uploaded documents doesn't work with the current configuration, so it should be safe to remove.


camillevilla:
👆 user error 😅

PR #1303, opened 2018-12-10
add honeybadger javascript reporting

aeschylus:


config/initializers/revision.rb
```diff
@@ -0,0 +1 @@
+Settings.revision = File.read("#{Rails.root}/REVISION").chomp if File.exists?("#{Rails.root}/REVISION")
```
aeschylus: Not sure this is how others would do it, but it seems to work on stage.

https://app.honeybadger.io/projects/49092/faults/42851601#notice-context was generated from called `window.onerror()` in the browser console.

app/views/layouts/spotlight/spotlight.html.erb
```diff
@@ -25,6 +25,15 @@
     <% end %>
     <%= stylesheet_link_tag "https://www.stanford.edu/su-identity/css/su-identity.css" %>
     <%= javascript_include_tag "application", async: Settings.async_javascript %>
+    <script type="text/javascript">
+     Honeybadger.configure({
+       apiKey: '<%= Honeybadger.config.get(:api_key) %>',
+       environment: '<%= Honeybadger.config.get(:env) %>',
+       debug: true,
+       onerror: true,
+       revision: 'git SHA/project version'
```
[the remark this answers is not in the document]
aeschylus: See comment in new commit.

config/initializers/revision.rb
jkeck: Maybe this could just go in `config/settings.yml` so there is no confusion as to where it comes from.

app/views/layouts/spotlight/spotlight.html.erb
```diff
@@ -25,6 +25,15 @@
     <% end %>
     <%= stylesheet_link_tag "https://www.stanford.edu/su-identity/css/su-identity.css" %>
     <%= javascript_include_tag "application", async: Settings.async_javascript %>
+    <script type="text/javascript">
+     Honeybadger.configure({
+       apiKey: '<%= Honeybadger.config.get(:api_key) %>',
+       environment: '<%= Honeybadger.config.get(:env) %>',
+       debug: true,
```
mejackreed: Should we output the debug info here? Has this been discussed already?

aeschylus: It hasn't been discussed yet. I think it should probably be turned off. I'll make the change.

PR #1309, opened 2018-12-14
Don't send invited users through Devise's password confirmation...

### LINKED ISSUE #1243: New exhibit users are being directed to set password once they accept email invite
This just happened to Dinah Handel. I sent Dinah an invite to the following exhibit:
https://exhibits.stanford.edu/jarndyce-test

Dinah received an email invite as expected. When she clicked on the button to accept the invite, she was presented with the following:

screen shot 2018-09-12 at 2 19 48 pm

Expected behavior:
New users should not have to create a special password to access their exhibit(s) -- all they should have to do is webauth in to exhibits.stanford.edu, and then they should see the "Your exhibits" tab.

jkeck:
... and pass them through our authentication service instead.

Closes #1243

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -14,9 +14,9 @@ class User < ApplicationRecord
   include Blacklight::User
 
   # Include default devise modules. Others available are:
-  # :confirmable, :lockable, :timeoutable and :omniauthable
+  # :confirmable, :lockable, :registerable, :recoverable, :timeoutable and :omniauthable
   devise :invitable, :remote_user_authenticatable, :database_authenticatable,
-         :registerable, :recoverable, :rememberable, :trackable, :validatable
+         :rememberable, :trackable, :validatable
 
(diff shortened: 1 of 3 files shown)


app/models/user.rb
```diff
@@ -14,9 +14,9 @@ class User < ApplicationRecord
   include Blacklight::User
 
   # Include default devise modules. Others available are:
-  # :confirmable, :lockable, :timeoutable and :omniauthable
+  # :confirmable, :lockable, :registerable, :recoverable, :timeoutable and :omniauthable
   devise :invitable, :remote_user_authenticatable, :database_authenticatable,
-         :registerable, :recoverable, :rememberable, :trackable, :validatable
+         :rememberable, :trackable, :validatable
```
jkeck: This didn't ultimately solve the problem, but seemed like a good idea?  `/shrug`

PR #1353, opened 2019-04-17
Remove unnecessary locale defaults; they don't seem necessary, and br…

### LINKED ISSUE #1349: Editing non-English homepage reverts to English version
This came up as a Vatican issue but I see it in Exhibits as well, but **not** with a fresh install of Spotlight, so I'm assuming the bug lies in Exhibits somehow (could be wrong, though).

In both Exhibits, and in both the http://vatican-dev.sul.stanford.edu and http://spotlight.bav.local instances:

1. Change to the Italian locale (I assume any non-English locale) 

2. Go to the homepage

3. Use the Edit button to edit the homepage 

You will then be in the English locale, editing the English version of the homepage rather than the Italian version as expected.

Note that the "Edit" link for the homepage on Curation > Translations > Pages page does the right thing; this seems specific to the Edit button link on the homepage itself.

cbeer:
…eak named URL routes. Fixes #1349

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -11,7 +11,7 @@ Exhibits::Application.routes.draw do
     mount Sidekiq::Web => '/sidekiq'
   end
 
-  scope '(:locale)', locale: Regexp.union(Spotlight::Engine.config.i18n_locales.keys.map(&:to_s)), defaults: { locale: nil } do
+  scope '(:locale)', locale: Regexp.union(Spotlight::Engine.config.i18n_locales.keys.map(&:to_s)) do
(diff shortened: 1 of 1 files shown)


jkeck:
This looks good to me from a code perspective.  I haven't tested the functionality yet though.

corylown:
We should determine if this is still needed.

This PR in exhibits fixed the issue also being addressed here: https://github.com/projectblacklight/spotlight/pull/2360
There's also an open PR to add this fix to spotlight: https://github.com/projectblacklight/spotlight/pull/2361

PR #1360, opened 2019-05-09
Swap chromedriver-helper for webdrivers

cbeer:


camillevilla:
Depends on https://github.com/rudionrails/yell/pull/54

PR #1378, opened 2019-05-23
Unpin sqlite and update dependencies

jkeck:
(plus bonus 🔪 ☕️-Script)


Gemfile.lock
```diff
@@ -83,7 +83,7 @@ GEM
       bootstrap-sass (~> 3.0)
       rails
       ruby-oembed
-    blacklight-spotlight (2.6.0)
+    blacklight-spotlight (2.6.1.1)
```
jkeck: Not sure if there is anything in this update that we want to account for.

mejackreed: Nah I think we are good. This fixes a bug

PR #1389, opened 2019-06-27
Add a link inside the full-text highlight section to initialize the...

jkeck:
...viewer on the record view with a prepared search.

Potentially closes sul-dlss/sul-embed#1022

forced-search

### DIFF
--- a/app/assets/javascripts/full_text_collapse.js
+++ b/app/assets/javascripts/full_text_collapse.js
@@ -14,10 +14,23 @@ Blacklight.onLoad(function(){
   });
 
   $('dt.blacklight-full_text_tesimv').each(function() {
-    $(this).text($(this).text().replace(/:$/, ''));
-    $(this).append('<span class="caret caret-right"></span>');
-    $(this).attr('data-toggle', 'collapse');
-    $(this).attr('data-target', '#' + $(this).next('dd').attr('id'));
+    var $dt = $(this);
+    var $dd = $dt.next('dd');
+    var $link = $dd.find('a.prepared-search-link');
+
+    $dt.text($dt.text().replace(/:$/, ''));
+    $dt.append('<span class="caret caret-right"></span>');
+    $dt.attr('data-toggle', 'collapse');
+    $dt.attr('data-target', '#' + $dd.attr('id'));
+
+    if ($link.length > 0) {
+      $dt.before(
+        $('<dt class="prepared-search-container"></dt>')
+          .html($link.clone())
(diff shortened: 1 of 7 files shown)


jkeck:
@jvine, I ended up putting the link inside the collapsible section due to some technical limitations w/ our current implementation.  We're using Blacklight's field configuration for the full-text field (so it plays nicely with the various Spotlight/Blacklight features).

This means that we're mainly constrained to customize the content in the `<dd>` for this particular field (w/o radically changing how the full-text feature works or implementing some pretty hacky workarounds).

I understand if this isn't acceptable, but thought I would put it out there as an easy alternative.

ggeisler:
@jkeck I'm happy for @jvine to think about this and answer whether this implementation is okay, but my initial reaction is that it is not really ideal. I understand why you took this approach but I'm worried there could commonly be cases where a user who would be interested in seeing what their query terms turn up in a given document would not necessarily make the effort to open the collapsed section, and so wouldn't be aware that prepared link is available.

Maybe it's a compromise we have to make, but just wanted to be clear that I do think it is better if that link is visible outside of the collapsed section.

jvine:
Agree that the link should be visible outside the collapsed section. I was nervous about creating a `<dt>` without an accompanying `<dd>` but I think for now we are going ahead with a blank `<dd>`. Link will appear directly above the toggle.

I did explore a bit whether the link could be in a `<dd>` below the snippets toggle `<dt>`, but visible and floated to the right. Maybe semantically better (?), but it does change the visual effect. Maybe in a good way? But may be tricky at narrower widths. 

Screen Shot 2019-06-27 at 10 43 19 AM

Screen Shot 2019-06-27 at 10 43 54 AM

jkeck:
prepared-search

app/assets/javascripts/full_text_collapse.js
```diff
@@ -14,10 +14,23 @@ Blacklight.onLoad(function(){
...
+    var $dd = $dt.next('dd');
+    var $link = $dd.find('a.prepared-search-link');
+
+    $dt.text($dt.text().replace(/:$/, ''));
+    $dt.append('<span class="caret caret-right"></span>');
+    $dt.attr('data-toggle', 'collapse');
+    $dt.attr('data-target', '#' + $dd.attr('id'));
+
+    if ($link.length > 0) {
+      $dt.before(
```
jkeck: Not super pleased about having to go down this route, but the alternative all seemed worse.

PR #1405, opened 2019-07-16
Always render the full text helper for documents that have full text …

### LINKED ISSUE #1398: Always display a link for "search for "search term" in document text" when a result has full text
Excerpted from @caaster; feel free to update. 

Re: full text search results in an exhibit that do not display snippets OR the "search for "search term" in document text" links

>Right now Solr is configured in such a way that we know the search term is present in the metadata OR the text for the item with no links. It is possible, though we haven't found an example yet, that the search term is ONLY in the metadata for the item.

>The short term solution that Jennifer has suggested, and Gary concurs with, is to always display a link for "search for "search term" in document text" -- when a search result has full text, of course -- until we have the bandwidth to research/implement other solution

jkeck:
…where there is a query passed in the search context.

This also changes the JS to remove the highlight section if it didn't contain any actual highlights (keeping the prepared search link intact)

Closes #1398

### DIFF
--- a/app/controllers/catalog_controller.rb
+++ b/app/controllers/catalog_controller.rb
@@ -226,7 +226,7 @@ class CatalogController < ApplicationController
         # This is required for the metadata configuration admin page to return the field properly.
         return true if args.length < 3
 
-        full_text_highlight_exists_in_response?(*args)
+        document_has_full_text_and_search_is_query?(*args)
       end,
       label: 'Sample matches in document text',
       highlight: true,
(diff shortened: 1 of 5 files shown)


app/controllers/catalog_controller.rb
```diff
@@ -393,18 +393,8 @@ def start_new_search_session?
...
-    def full_text_highlight_exists_in_response?(context, _config, document)
-      response = context.instance_variable_get(:@response) || {}
-      document_highlight = response.dig('highlighting', document['id'])
-      return true if document_highlight.present? && document_highlight.any? do |field, values|
-        Settings.full_text_highlight.fields.include?(field) && values.present?
-      end
-
-      false
+    def document_has_full_text_and_search_is_query?(context, _config, document)
+      context.params[:q].present? && document['full_text_tesimv'].present?
```
cbeer: Does this mean we have to return the whole `full_text_tesimv` field for every request? Maybe we'll need to introduce a new field (or look into some solr-side magic) to avoid that.

jkeck: I was concerned/worried about the same thing, but it appears to work?

The search results in our app appear to know if the `full_text_tesimv` is present or not (which was surprising after looking at the solr configs).  I guess it's possible that our local app configs don't match what's on the server, but that's the behavior that I'm getting in development. 

My other thought was to add a boolean field into the document, but not having to modify the indexing seemed enticing.

jkeck:
I'm going to deploy this to stage to ensure that this works in a production context (and isn't something odd going on w/ our dev configs)

jkeck:
It appears to work in stage as well (using the vt-test exhibit as an example)

camillevilla:
I've left this unmerged so that @caaster can do the final review when she returns to the office.

PR #1418, opened 2019-08-19
Update dependencies

sul-devops-team:


Gemfile.lock
```diff
@@ -568,21 +568,20 @@ GEM
...
     selenium-webdriver (3.142.3)
       childprocess (>= 0.5, < 2.0)
       rubyzip (~> 1.2, >= 1.2.2)
-    sidekiq (5.2.7)
-      connection_pool (~> 2.2, >= 2.2.2)
-      rack (>= 1.5.0)
-      rack-protection (>= 1.5.0)
-      redis (>= 3.3.5, < 5)
-    sidekiq-pro (4.0.5)
+    sidekiq (6.0.0)
```
mejackreed: We may want to be weary about this also. https://github.com/mperham/sidekiq/blob/master/Changes.md#60

camillevilla: Agreed. Let's work on that in a separate PR next week.

PR #1447, opened 2019-11-07
Update i18n fallbacks config

camillevilla:
Update i18n fallbacks setting per `i18n` [v1.1.0](https://github.com/ruby-i18n/i18n/releases/tag/v1.1.0). 
Also see: https://github.com/ruby-i18n/i18n/wiki/Fallbacks

### DIFF
--- a/config/environments/development.rb
+++ b/config/environments/development.rb
@@ -58,6 +58,10 @@ Rails.application.configure do
   # Raises error for missing translations
   # config.action_view.raise_on_missing_translations = true
 
+  # Enable locale fallbacks for I18n (makes lookups for any locale fall back to
+  # the I18n.default_locale when a translation cannot be found).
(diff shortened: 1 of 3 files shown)


config/environments/development.rb
```diff
@@ -58,6 +58,10 @@
   # Raises error for missing translations
   # config.action_view.raise_on_missing_translations = true
 
+  # Enable locale fallbacks for I18n (makes lookups for any locale fall back to
+  # the I18n.default_locale when a translation cannot be found).
+  # config.i18n.fallbacks = [I18n.default_locale]
```
camillevilla: Adding these as commented-out options because I spent a second scratching my head over why I was getting this behavior in dev 😅
fallbacks disabled; 'translation missing'

PR #1448, opened 2019-11-07
Set HTML lang to i18n locale

camillevilla:
There's probably more due diligence to be done here (see [W3C: Authoring HTML: Language declarations](https://www.w3.org/TR/i18n-html-tech-lang/)), but I would expect to discuss some of the hairier issues in upstream Spotlight and DLME work (i.e. how should the markup change when one is doing curation in one language but previewing text that will displayed in a different langauge, etc.)


camillevilla:
There's no real rush to ship this, so I guess we can wait until we have access to projectblacklight/blacklight#2203?

Issue #1456, opened 2019-11-27
Private item displays to all users in newly created Browse category for published exhibit


caaster:
Amy Hodge created the following new browse category in exhibits:
https://exhibits.stanford.edu/data/browse/shockgas-ir

The browse category includes one item which Amy marked as private.

The private item is visible by any exhibit user (not logged in).

If this bug ends up being pervasive, it could be quite problematic -- exposing items to the public that should not be visible. Therefore, it would be good to have it looked into soon.

TODO:

- [x] add a uniqueness constraint to stop multiple sidecars from showing up (#1457)
- [x] remediate bad data
- [x] stop exhibits index statuses  from creating bad  data  (#1457)
- [x] stop needless exhibits indexing from happening (#1459)

cbeer:
It looks like we're getting some duplicate data in the database. I've removed the duplicates for that  particular record so it is now properly hidden.

PR #1457, opened 2019-12-06
Refactor RecordIndexStatusJob to avoid a race condition

cbeer:
Should address the underlying cause of #1456


cbeer:
Closed in favor of #1459

PR #1494, opened 2020-01-15
Add a simple API to query for exhibits that contain a given document …

jkeck:
…(or document in a given collection)

### DIFF
--- a/app/controllers/exhibit_finder_controller.rb
+++ b/app/controllers/exhibit_finder_controller.rb
@@ -1,0 +1,14 @@
+# frozen_string_literal: true
+
+##
+# This is a simple controller to return the ExhibitFinder JSON response for a given ID (druid)
+class ExhibitFinderController < ApplicationController
+  before_action do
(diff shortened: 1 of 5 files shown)


config/routes.rb
```diff
@@ -11,6 +11,8 @@
     mount Sidekiq::Web => '/sidekiq'
   end
 
+  resources :exhibit_finder, only: :show
```
jkeck: I could feature flag this route if we wanted to keep it out of prod, but it seems relatively harmless.

app/controllers/exhibit_finder_controller.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+##
+# This is a simple controller to return the ExhibitFinder JSON response for a given ID (druid)
+class ExhibitFinderController < ApplicationController
+  before_action do
+    headers['Access-Control-Allow-Origin'] = '*'
```
mejackreed: 👌 🤷‍♂

PR #1523, opened 2020-01-17
Add icon labels for view icons fixes #1515

### LINKED ISSUE #1515: BL7 upgrade: Search type icon titles have "icon" in them
Not sure if this is an upstream spotlight issue, or local, but creating the ticket here for now.

Screen Shot 2020-01-16 at 10 04 30 AM

mejackreed:
Screen Shot 2020-01-16 at 5 38 06 PM


cbeer:
Is this something we should also be porting upstream?

mejackreed:
I thought about that.. seems like we could. But we would have to do some of this anyways since Spotlight doesn't include all of these views by default. /shrug

PR #1536, opened 2020-01-18
Allow the maxheight of embed in the item embed widget to be set

### LINKED ISSUE #1196: Change size/display of sul-embed frame for item embed widget
The exhibit creator for the Baltic exhibit recently filed feedback about this. Her workaround was to feature her videos from youtube, instead of those same videos from the SDR. We really don't want this to happen! Here are screenshots of the youtube widget & item embed widget for her exhibit.

screen shot 2018-06-07 at 10 01 49 am

screen shot 2018-06-07 at 10 03 22 am

jkeck:
Closes #1196 

resize-item-embed

### DIFF
--- a/app/assets/javascripts/sir_trevor_block_overrides.js
+++ b/app/assets/javascripts/sir_trevor_block_overrides.js
@@ -1,0 +1,10 @@
+SirTrevor.Blocks.SolrDocumentsEmbed = (function(){
+  return SirTrevor.Blocks.SolrDocumentsEmbed.extend({
+    item_options: function() {
+      return [
+        '<label for="<%= formId("maxheight") %>">Maximum height of viewer (in pixels)</label>',
+        '<input id="<%= formId("maxheight") %>" type="number" class="form-control" placeholder="600" name="maxheight" />'
+      ].join(' ');
+    }
+  });
+})();
--- a/spec/features/embed_block_spec.rb
+++ b/spec/features/embed_block_spec.rb
@@ -1,0 +1,51 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+# rubocop:disable Rails/DynamicFindBy
+# Gem::Specificationfind_by_name is not a rails dynamic finder
+require "#{Gem::Specification.find_by_name('blacklight-spotlight').gem_dir}/spec/support/features/test_features_helpers"
+# rubocop:enable Rails/DynamicFindBy
+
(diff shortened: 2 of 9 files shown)


app/assets/javascripts/sir_trevor_block_overrides.js
```diff
@@ -0,0 +1,10 @@
+SirTrevor.Blocks.SolrDocumentsEmbed = (function(){
+  return SirTrevor.Blocks.SolrDocumentsEmbed.extend({
+    item_options: function() {
+      return [
+        '<label for="<%= formId("maxheight") %>">Viewer maximum height (in pixels)</label>',
```
jkeck: If this seems like a reasonable approach, @ggeisler I'm assuming we'll want to change this label.

ggeisler: Sure, this looks nice @jkeck. I assume there is a default value that is visible in the input when the user first edits it (like the `600` in your gif)? If so, I think that is a good enough clue for the user to understand they should just enter a number (otherwise more CSS-savvy users might not know whether to enter `300` or 300px`).

I'd suggest a very minor update to the label:

> Maximum height of viewer (in pixels)

jkeck: @ggeisler yeah, I'm using `600` as a `placeholder` in that element (which is the default if nothing is set, and is what we're currently doing so this should be all backwards compatible with existing pages).

I'm also using a `number` field so users will only be able to enter numeric data (which will also hopefully help eliminate any ambiguity about what data to enter).

I'll update that label, think about how to test this appropriately, and get it ready for review.  Thank you!

camillevilla:
Neat work @jkeck. I took this for a test drive with a media player item and took some screenshots.

### 600px
600px
### 250px
250px

spec/features/embed_block_spec.rb
```diff
@@ -0,0 +1,51 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+# rubocop:disable Rails/DynamicFindBy
+# Gem::Specificationfind_by_name is not a rails dynamic finder
+require "#{Gem::Specification.find_by_name('blacklight-spotlight').gem_dir}/spec/support/features/test_features_helpers"
```
jkeck: 😬

Any better way to do this?

camillevilla: ¯\_(ツ)_/¯

PR #1581, opened 2020-01-30
Update tag styling fixes #1573

### LINKED ISSUE #1573: BL7 upgrade: update styling of exhibit tags on landing page
Just a few minor CSS updates. For reference, here's what's on -stage:

Screen Shot 2020-01-29 at 5 31 29 PM

---

Here's the goal:

Screen Shot 2020-01-29 at 5 28 33 PM

---

**A:** The background color of the currently selected tag is different from what's on -prod, but I think this is okay as-is. Since we'll be adding the search across elements to this page, I think it might be better that the selected tag is not using the primary red color (since tags are no longer the primary thing a user might want to select on the page).

The alignment of the buttons also changed to centered from left-aligned, but again, I think that is fine as-is.

- [x] **B:** I'd like to tighten the tags up a bit (to fit more horizontally and generally make them feel more concise and clean). Here's what I suggest:

```SCSS
.tags .nav-link {
    padding: 0.33rem 0.66rem;
}
```

- [x] **C:** When the tags wrap, we need some space between rows. I tried using (in Chrome) the `grid-gap` property (and some variants) but that didn't seem to do anything. I'm rusty on Flexbox so maybe I was doing it incorrectly, or maybe it isn't something supported in Chrome yet. So if we can't get a gap between rows of tags using a flex property, a fallback might be to add bottom margin to the selector above:

```SCSS
.tags .nav-link {
    padding: 0.33rem 0.66rem; // for more compact buttons
    margin-bottom: 0.5rem; // for row spacing
}
```
That's what I did to get the result in the screenshot above (though note I also changed the font-size back to `15px`, which will come in a separate PR or ticket).

On a smaller screen, it should look something like:

Screen Shot 2020-01-29 at 5 29 47 PM

mejackreed:
Screen Shot 2020-01-30 at 3 23 15 PM

### DIFF
--- a/app/assets/stylesheets/modules/spotlight_overrides.scss
+++ b/app/assets/stylesheets/modules/spotlight_overrides.scss
@@ -196,4 +196,10 @@ h5,
 .img-thumbnail > a > img {
   @include thumbnail-image-border;
 }
+
+.tags .nav-link {
+  margin-bottom: 0.5rem;
+  padding: 0.33rem 0.66rem;
+}
+
 // scss-lint:enable ImportantRule


app/assets/stylesheets/modules/spotlight_overrides.scss
```diff
@@ -195,3 +195,8 @@ h5,
 .img-thumbnail > a > img {
   @include thumbnail-image-border;
 }
+
+.tags .nav-link {
+  margin-bottom: 0.5rem;
+  padding: 0.33rem 0.66rem;
```
jkeck: Is this something that should be upstream or is this specific to our app?

mejackreed: I talked with @ggeisler about this. His recommendation was that Spotlight should continue using the BS4 nav pills and this custom styling should reside in Exhibits. Exhibits uses a different approach to styling the nav pills than BS4 (pills are button looking vs BS4 only active looks like a button)

jkeck: gotcha

Issue #1549, opened 2020-01-24
BL7 upgrade: adjust form field label alignment on Admin pages


ggeisler:
On all forms with field labels in the Admin pages the labels are currently left-aligned and unstyled. They should be right-aligned and bolded, as they currently are in production:

<img width="1084" alt="Screen Shot 2020-01-24 at 9 47 46 AM" src="https://user-images.githubusercontent.com/101482/73087039-0ac2b780-3e8f-11ea-9cff-95cf17a54cbb.png">

I was going to list all of the forms this affects, but I think it is every form with a field label in the Admin section, including forms that don't have text inputs but still have form field labels:

<img width="818" alt="Screen Shot 2020-01-24 at 9 54 16 AM" src="https://user-images.githubusercontent.com/101482/73087342-9d635680-3e8f-11ea-9c00-afb95dafbbbf.png">

jkeck:
This is a change that was made in bootstrap between versions 3 and 4.

## Bootstrap 3
<img width="877" alt="bs-3 3" src="https://user-images.githubusercontent.com/96776/73295265-365dde80-41bc-11ea-9bd7-b4256a866eeb.png">

## Bootstrap 4
<img width="820" alt="bs-4 1" src="https://user-images.githubusercontent.com/96776/73295283-3eb61980-41bc-11ea-8162-dc2da9f4093d.png">

If we do want to retain the previous styling, then we need to decide if this is a change we should be making purely for ourselves our is a decision we want to make upstream in Spotlight.

ggeisler:
After discussion with JV, we want to stick to the suggestions in my original comment of this ticket.

In other words, despite the change made by BS4, we want to right-align and bold field labels as we currently do in production with BS3.

jkeck:
@ggeisler and should this be just a local customization in Exhibits or is the suggestion that we do this upstream in Spotlight?

ggeisler:
@jkeck Yeah, I'm really not sure how to make that decision. JV and I agreed we think for the type of application Spotlight is we think the BS3 way of styling the field labels was better.

Would everyone agree with us? Apparently not BS4, so I don't know. (I still think it was an odd change for them to make.)

But we've made a lot of design decisions for Spotlight before, and since JV agreed with me it's not just a weird preference of mine, so I guess I'm suggesting, yes, do it in Spotlight so everyone gets the old, better approach.

I wouldn't object if anyone sees it differently, though.

PR #1599, opened 2020-01-31
form-group to form-row

### LINKED ISSUE #1594: BL7 upgrade: make Configuration > Viewers label inline
I think the Viewers form label would look better (and be consistent with our other Admin page forms) if it were inline with the other form elements:

Screen Shot 2020-01-31 at 12 23 10 PM

(I also want to add some vertical space between options, but that's a global issue I'll deal with in a separate vertical spacing PR or ticket.)

mejackreed:
Fixes #1594 

@ggeisler this doesn't keep the exact experience but since some of that was removed in BS4 I'm wondering if something like this works?

Screen Shot 2020-01-31 at 3 00 09 PM


ggeisler:
@mejackreed I think that will be fine after we implement #1549 to make the labels right-aligned.

If we left it like this I think it would be less than ideal because of the distance between the label and the options, but that's what #1549 aims to address.

PR #1638, opened 2020-02-07
Add a start_over_html i18n key that includes a "Back" svg.

### LINKED ISSUE #1610: BL7 upgrade: add icon to Start over button on results page
As part of the BL7 upgrade the search results header controls changed a bit, moving the Start Over and search constraints to left-align with the sidebar rather than the main content area as currently seen in -prod.

Overall this change is fine for the search results page. (We would like to make a change to the item detail page, which I'll [ticket separately](https://github.com/sul-dlss/exhibits/issues/1611).) But to follow the SearchWorks pattern a bit more closely, and maybe make it slightly more clear to the user that "Start Over" returns them to the homepage, it would be great if we could:

- [x] add an icon to the "Start Over" button, similar to SW

Screen Shot 2020-02-03 at 4 22 27 PM

In the mockup above I've used the "FirstPageSharp" icon from the Material UI icon set (https://material-ui.com/components/material-icons/) (apparently you can no longer provide a direct link to a MUI icon...).

- [x] To be more consistent with SUL's sentence-case writing style, it would be great to update the label to read "Start over" rather than "Start Over", unless that is technically difficult.

jkeck:
If this seems like an 👌 approach, then this would need a release that includes projectblacklight/blacklight#2238 in order to work.

Closes #1610 

Screen Shot 2020-02-07 at 2 14 44 PM

### DIFF
--- a/app/assets/images/blacklight/start_over.svg
+++ b/app/assets/images/blacklight/start_over.svg
@@ -1,0 +1,3 @@
+<svg xmlns="http://www.w3.org/2000/svg" width="24" height="24" viewBox="0 0 24 24" aria-hidden="true" role="presentation">
+  <path d="M18.41 16.59L13.82 12l4.59-4.59L17 6l-6 6 6 6 1.41-1.41zM6 6h2v12H6V6z"></path>
+</svg>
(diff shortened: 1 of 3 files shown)


cbeer:
I was looking at #1610 earlier (and gave up on it) after having trouble getting the button back to the right size and getting the icon vertically aligned correctly.

ggeisler:
@jkeck Obviously I can't speak to the technical complications of doing this, but that looks great.

jkeck:
After discussion, the icon alignment appears to be a systemic issue, and shouldn't necessarily hold this up.

I'm marking this ready for review as this will update the text currently, and once the partial support is available in the release of Blacklight we're on the icon will become available.

PR #1646, opened 2020-02-12
Fixes an issue where ordering of ControllerOverride caused the RangeL…

mejackreed:
…imit has_search_params? method to not be called

### DIFF
--- a/app/controllers/catalog_controller.rb
+++ b/app/controllers/catalog_controller.rb
@@ -3,11 +3,11 @@
 ##
 # Blacklight controller providing search and discovery features
 class CatalogController < ApplicationController
-  include BlacklightRangeLimit::ControllerOverride
   include BlacklightAdvancedSearch::Controller
   helper Openseadragon::OpenseadragonHelper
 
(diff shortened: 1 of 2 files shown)


app/controllers/catalog_controller.rb
```diff
@@ -3,11 +3,11 @@
 ##
 # Blacklight controller providing search and discovery features
 class CatalogController < ApplicationController
-  include BlacklightRangeLimit::ControllerOverride
   include BlacklightAdvancedSearch::Controller
   helper Openseadragon::OpenseadragonHelper
 
   include Blacklight::Catalog
+  include BlacklightRangeLimit::ControllerOverride
```
jkeck: 👍 are other folks going to need to ensure that this include comes after `Blacklight::Catalog`?  Just wondering if we should spin off a ticket about updating the generator to be more intelligent about where it's injecting this include.

mejackreed: Great idea, added https://github.com/projectblacklight/blacklight_range_limit/issues/134

PR #1678, opened 2020-02-14
Revise font sizes

ggeisler:
Nothing too major, but sorts out some inconsistency in sizing between the H1 and H2 sizes, increases the small help text size a bit, and adjusts the masthead title and subtitle sizes. I also did a little responsive sizing to the masthead titles, just because below `lg` the title feels bigger than it should be given the reduced width of the viewport.

Will double-check everything again when it gets to -stage.

Screen Shot 2020-02-14 at 11 33 11 AM

### DIFF
--- a/app/assets/stylesheets/modules/spotlight_overrides.scss
+++ b/app/assets/stylesheets/modules/spotlight_overrides.scss
@@ -87,7 +87,7 @@
   font-size: 1.25rem;
 }
 
-.tab-pane small.form-text {
+small.form-text {
   font-size: 0.875rem;
 }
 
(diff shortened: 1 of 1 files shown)


app/assets/stylesheets/modules/spotlight_overrides.scss
```diff
@@ -40,22 +67,45 @@
...
+.blacklight-browse h1 {
+  font-size: 2rem;
+}
+
 .item-count {
   color: $cool-grey;
   font-size: 1.25rem;
 }
 
+.tab-pane small.form-text {
```
jkeck: Just thinking of the change I made in projectblacklight/spotlight#2465, I'm wondering if this should not be constrained to only `small.form-text` elements that are children of a `.tab-pane`?

ggeisler: Right, I saw that change you made and you're right it probably makes sense to do here. I'll make the update now.

Issue #1661, opened 2020-02-13
Need toggle affordance on Curation > Add items page


ggeisler:
The Item status and Item submission panels on the Curation > Add items page used to have visible toggle icons:

![Screen Shot 2020-02-12 at 5 03 46 PM](https://user-images.githubusercontent.com/101482/74391696-d4a39400-4db9-11ea-874c-c9ba468988bd.png)

But these have disappeared in the BL7 upgrade:

![Screen Shot 2020-02-12 at 5 03 26 PM](https://user-images.githubusercontent.com/101482/74391720-e2f1b000-4db9-11ea-991f-cab985871853.png)

I think we need visible toggle icons because I don't think the panel title being a link is a strong enough affordance for all users to understand these panels can be opened and closed.

mejackreed:
Resolved in https://github.com/sul-dlss/exhibits/pull/1673

mejackreed:
Reopening

PR #1692, opened 2020-02-18
Temporary fix so that items can be added, but still needs work

mejackreed:
`.stretched-link` is too aggressive here, causing issues where items cannot be added to sdr block. Additional work needs to be done so that the entire button area is clickable

Reopens #1661 (not a blocker)


cbeer:
I think I ran into this before; does the `h4` or `card-header` need `position`?

cbeer:
See #1694

mejackreed:
👍 closing in favor of #1694

PR #1711, opened 2020-02-19
Remove hover / focus colors from topnav links

### LINKED ISSUE #1647: Remove :focus and :hover styling from topbar links
It looks like our topbar links get some default BS4 styling applied for the `:hover` and `:focus` states:

`..navbar-dark .navbar-nav .nav-link:hover, .navbar-dark .navbar-nav .nav-link:focus` -> `color: rgba(255,255,255,0.75);`

This results in the link text being a more faint white in those states (more obvious in action than in the screenshot):

Screen Shot 2020-02-12 at 12 45 40 PM

I suggest we remove that styling so that there is no change to the font color in those states; we can rely on the existing `text-decoration: underline` for the hover state to indicate to the user that they are hovering on this link.

camillevilla:
Closes #1647

hover-link-styles

### DIFF
--- a/app/assets/stylesheets/modules/spotlight_overrides.scss
+++ b/app/assets/stylesheets/modules/spotlight_overrides.scss
@@ -1,5 +1,8 @@
-#header-navbar {
-  a {
+// Override Bootstrap 4 nav-dark link states
+#user-util-collapse {
+   :not(.dropdown-menu) > li > a {
+    color: $white;
+
     &:hover,
     &:focus,
     &:active {


app/assets/stylesheets/modules/spotlight_overrides.scss
```diff
@@ -1,5 +1,8 @@
-#header-navbar {
-  a {
+// Override Bootstrap 4 nav-dark link states
```
mejackreed: Thank you for this helpful comment

PR #1722, opened 2020-02-20
Add two high-level search across tests that validate that...

### LINKED ISSUE #1716: Backfill some broad tests for search across

jkeck:
...searching is constrained to exhibits the user has access to see

Closes #1716


jkeck:
There is the opportunity to add more tests (and updates to fixtures to support them) around the public/private aspect of documents themselves, but those can possibly grow as we  build out the feature.

PR #1729, opened 2020-02-21
Update home page card font sizes

### LINKED ISSUE #1728: Home page cards - Font sizes are inaccurate.
The 3 elements of the card are meant to be 3 different font sizes. The proposed font sizes were:

> title `<h2>` has font-size: 1.125rem;
> subtitle has font-size: 1rem;
> description has font-size: .8rem; with line-height: 1.6em;

Description currently has .875rem, and the line-height has not been set. Title `<h2>` is smaller than intended.

Looks like these font adjustments may affect the truncation further, but it's important that these three elements indicate a hierarchy.

cbeer:
Fixes #1728 

Screen Shot 2020-02-21 at 10 06 13

### DIFF
--- a/app/assets/stylesheets/modules/home.scss
+++ b/app/assets/stylesheets/modules/home.scss
@@ -31,11 +31,6 @@
   }
 }
 
-.card-front img {
-  border-bottom: 1px solid $color-beige-40;
-  box-shadow: 0 4px 6px -6px $color-gray-60;
-}
-
 .new_user .form-actions {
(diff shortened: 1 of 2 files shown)


jvine:
Font and truncation look good in the context of stage; just noticed this one oddity that the slide-up is different for hover vs. focus.

mejackreed:
@jvine should be resolved in #1732 ... but let us know if not.

PR #1731, opened 2020-02-21
Remove old gdor config

jcoyne:
The gdor indexer that used this was removed in #847


.gitignore
```diff
@@ -18,7 +18,6 @@
 # Ignore ruby version management artifacts.
 /.ruby-*
 
-/config/gdor.yml
 /log/gdor_logs/
```
camillevilla: Should these be removed as well?

jcoyne: no, we've repurposed this directory for harvestdor logs

PR #1734, opened 2020-02-24
Override Spotlight partial so we can have responsive sidebar locations

mejackreed:


app/views/spotlight/exhibits/index.html.erb
```diff
@@ -0,0 +1,51 @@
...
+
+  <% if current_user && current_user.exhibits.any? %>
+    <div role="tabpanel" class="tab-pane" id="user">
+      <%= render 'exhibits', exhibits: current_user.exhibits %>
+    </div>
+  <% end %>
+</div>
+
+<% # Overridden just to override these classes %>
+<% content_for(:content_position) { 'order-12 order-lg-0' } %>
```
mejackreed: Should we just do this in Spotlight?

jkeck: I think seeing something upstream in Spotlight would be interesting for this.  It certainly doesn't work super well as-is, so any improvement would be 👍 from me.

cbeer:
I think the new design makes this unnecessary.

PR #1739, opened 2020-02-24
Use post methods to query for large document sets fixes #1737

### LINKED ISSUE #1737: Reference item show page fails to load when many documents
It's possible this is a problem with my browser setup, but I see the same issue on both Chrome and Firefox so it's at least worth someone else verifying.

It appears that for Parker reference items that are included in many different Parker documents, the list of documents is too large to load, so we just get the "This reference is found in these document bibliographies" heading but the list of documents that should appear below does not ever show up.

For example, I go to: https://exhibits.stanford.edu/parker/catalog/KAGI2SI5

No list of documents is displayed, but looking at the inspector there seems like a long list that should be displayed, and looking at the console (in Chrome, don't see the equivalent in Firefox) it seems like Chrome just gave up because the request was too large:

Screen Shot 2020-02-24 at 1 34 57 PM

mejackreed:
I'd like to try this in stage, but think it should work.

### DIFF
--- a/app/controllers/catalog_controller.rb
+++ b/app/controllers/catalog_controller.rb
@@ -25,6 +25,7 @@ class CatalogController < ApplicationController
   end
 
   configure_blacklight do |config|
+    config.http_method = :post
     config.index.default_thumbnail = :exhibits_default_thumbnail
 
     # default advanced config values
--- a/app/assets/javascripts/cited_documents.js
+++ b/app/assets/javascripts/cited_documents.js
@@ -11,7 +11,7 @@
       var data = $el.data();
       var solrQueryString = data.documentids.join(' OR ');
 
-      $.getJSON(data.path, {
+      $.post(data.path, {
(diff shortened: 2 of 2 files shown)


app/controllers/catalog_controller.rb
```diff
@@ -25,6 +25,7 @@ class CatalogController < ApplicationController
   end
 
   configure_blacklight do |config|
+    config.http_method = :post
```
mejackreed: Any performance implications or other side effects we should be concerned about here?

cbeer: It makes debugging using Solr log files harder 🤸‍♂

mejackreed: 😢 seems like this should have been broken for year(s) already. I'm curious if thats worth the trade off or we have other alternatives? (batched requests? 😢 )

Issue #1783, opened 2020-02-27
Site landing page searchbox sizing and menu alignment


ggeisler:
This is based on latest `master` running locally, so we can close if these are known issues that just haven't been completely addressed yet.

- [x] Searchbox in site menu isn't wide enough at `md` and larger viewports to contain the placeholder text (and there is space for it to be wider in those viewport widths)

- [x] The right-side menus are aligned more like left-side menus, in that when open they extend to the right instead of the left, thus cutting off the menu options at some viewport widths.

<img width="1184" alt="Screen Shot 2020-02-27 at 10 13 24 AM" src="https://user-images.githubusercontent.com/101482/75468624-bce81600-594a-11ea-934d-b4ba31052567.png">

cbeer:
What browser + viewport size are  you seeing that search box issue?

ggeisler:
Chrome and the screenshot is `xl`. But I believe it was also on `lg` (i.e., all viewports above where it collapses and takes full-width).

If you aren't seeing it I'll take another look after I switch back to running exhibits locally.

ggeisler:
@cbeer Yeah, it is only in Chrome (Version 80.0.3987.122 (Official Build) (64-bit)), at `md` and above. Below is what I currently see in Safari, Firefox, and Chrome (left to right):

<img width="1188" alt="Screen Shot 2020-02-27 at 4 22 49 PM" src="https://user-images.githubusercontent.com/101482/75496153-998a8e80-597d-11ea-821c-3a26be058344.png">

cbeer:
Firefox and Safari were fine with it. Ugh. We're not  currently setting  a width and expected the browser to flexbox appropriately. Can we just add, say, `min-width: 17em` and call it  a day?

ggeisler:
Sure, I was going to say, we only need a little bit more to work for Chrome so we just add a bit it should still look fine on the other browsers. `17em` looks good.

PR #1794, opened 2020-02-27
Set a min-width on the search box (for Chrome..)

cbeer:
part  of #1783

After:
Screen Shot 2020-02-27 at 15 33 52

### DIFF
--- a/app/assets/stylesheets/modules/layout.scss
+++ b/app/assets/stylesheets/modules/layout.scss
@@ -30,7 +30,14 @@
   }
 }
 
-.exhibit-navbar {
+.site-navbar {
+  .search-query-form {
+    min-width: 17em;
+  }
+}
+
+.exhibit-navbar,
+.site-navbar {
   .search-query-form {
     .btn-primary {
       @extend .btn-secondary;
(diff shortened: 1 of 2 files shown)


app/assets/stylesheets/modules/layout.scss
```diff
@@ -30,6 +30,12 @@
...
 
+.site-navbar {
+  .search-query-form {
+    min-width: 17em;
+  }
+}
+
 .exhibit-navbar {
   .search-query-form {
     .btn-primary {
```
camillevilla: A side-effect of this class name change is that this button is now cardinal red instead of the secondary beige-y color. Not sure if this is intentional?

## Before
Screen Shot 2020-02-27 at 4 13 54 PM

## After
Screen Shot 2020-02-27 at 4 33 41 PM

ggeisler: I'd lean toward the beige-y color, just because the red is pretty eye-catching and in this context I think browsing the cards is still the most expected user action.

cbeer: Thanks, nice catch. I think we still should update the class name, but have it share that style (as in 7bb380f)?

PR #1827, opened 2020-03-03
Only render the first search field (all/default) as a hidden input …

### LINKED ISSUE #1825: Fielded search option showing when in site admin pages
Unless this is a stubborn caching issue, when you are on one of the site admin pages, the fielded search option is shown, even though we decided to remove it for search across.

Screen Shot 2020-03-03 at 12 04 11 PM

jkeck:
… in the Search Across form.

Closes #1825 

## Site Users Admin (before)
admin-users-before

## Site Users Admin (after)
admin-users-after

This also affects a few other pages as well.

### DIFF
--- a/app/views/shared/_site_search_form.html.erb
+++ b/app/views/shared/_site_search_form.html.erb
@@ -4,18 +4,10 @@
     <label for="search_field" class="sr-only"><%= t('blacklight.search.form.search_field.label') %></label>
   <% end %>
   <div class="input-group">
-    <% if search_fields.length > 1 %>
-        <%= select_tag(:search_field,
-                       options_for_select(search_fields, h(params[:search_field])),
-                       title: t('blacklight.search.form.search_field.title'),
-                       id: "search_field",
-                       class: "custom-select search-field") %>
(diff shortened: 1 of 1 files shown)


app/views/shared/_site_search_form.html.erb
```diff
@@ -4,15 +4,7 @@
...
-    <% if search_fields.length > 1 %>
-        <%= select_tag(:search_field,
-                       options_for_select(search_fields, h(params[:search_field])),
-                       title: t('blacklight.search.form.search_field.title'),
-                       id: "search_field",
-                       class: "custom-select search-field") %>
-    <% elsif search_fields.length == 1 %>
-      <%= hidden_field_tag :search_field, search_fields.first.last %>
-    <% end %>
+    <%= hidden_field_tag :search_field, search_fields.first.last %>
```
jkeck: This means that if we decide to re-add a search fields drop-down we'll want to pull this back. If we do that we would need to figure out how to inject our search fields configurations into various Spotlight controller contexts, which seems like a recipe for disaster (so this seems like a simple solution for now).

PR #1831, opened 2020-03-04
Ensure that map does not break the metadata layout fixes #1787

### LINKED ISSUE #1787: Heatmap breaks metadata layout
The hidden `<dt>` for the heatmap seems to be breaking the layout for the remainder of the metadata.

I think there's an argument for displaying a label for the heatmap, since there's nothing to announce it to a SR. (Okay, the a11y label is tenuous, but I'm adding it anyway.)

https://exhibits.stanford.edu/radarfilm/catalog/jt744vh3080

**Current**
Screen Shot 2020-02-27 at 10 14 27 AM

**Label displayed**
Screen Shot 2020-02-27 at 10 14 42 AM

mejackreed:
Some Gifs of configurations that we use in production (top and bottom)

## Configured at the top
Kapture 2020-03-04 at 11 40 30

## Configured at the bottom
Kapture 2020-03-04 at 11 38 54

### DIFF
--- a/app/assets/stylesheets/modules/blacklight_heatmaps_overrides.scss
+++ b/app/assets/stylesheets/modules/blacklight_heatmaps_overrides.scss
@@ -1,9 +1,3 @@
-.blacklight-heatmaps-show-map-wrapper {
-  @media (min-width: breakpoint-min("sm")) {
-    padding-left: 50px;
-  }
-}
-
 .blacklight-heatmaps-show-map {
   margin-bottom: $spacer;
   width: 100%;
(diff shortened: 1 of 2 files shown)


ggeisler:
I know this PR is already merged and it looks great, just want to note that I'm not sure the label ("Inline Map") is ideal. I know the exhibit curator can override the default label and change it that way, but relatively few curators change field labels. So I'm wondering if it is possible to change the default so if the curator doesn't change it, it is a little more informative ("Inline Map" feels to me more like a backend description than one intended for end-users)?

I'm not sure this is the best, but something like "Geographic context" feels more end-user oriented to me.

PR #1844, opened 2020-03-12
Add the thumb field for IIIF imported items

mejackreed:
Adds support for a few cases so the results views look consistent:

 1. No thumbnail is provided (falls back to first canvas) https://cdm17229.contentdm.oclc.org/iiif/info/p17229coll19/192/manifest.json
 2. Thumbnail is provided https://purl.stanford.edu/bb389yg4719/iiif/manifest
Screen Shot 2020-03-12 at 5 04 26 PM

## Update

Adds a commit that removes the `square` region parameter. This is not supported by existing Exhibits content that @camillevilla pointed out. So while this is suboptimal, its better than broken images.

Screen Shot 2020-03-13 at 4 08 05 PM

### DIFF
--- a/app/models/spotlight/resources/iiif_manifest.rb
+++ b/app/models/spotlight/resources/iiif_manifest.rb
@@ -1,0 +1,25 @@
+# frozen_string_literal: true
+
+# This unpleasantness allows us to include the upstream class before overriding it
+# rubocop:disable Rails/DynamicFindBy
+spotlight_path = Gem::Specification.find_by_name('blacklight-spotlight').full_gem_path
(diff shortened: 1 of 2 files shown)


mejackreed:
Re: push upstream.

Kind of? Spotlight doesn't use the concept of `square` thumbs so that field would not be used. So we would need to maintain this patch also. And per this https://github.com/projectblacklight/spotlight/issues/2365#issuecomment-576760195 I don't know how much we would want to do with that. I think the idea being you should adapt and customize your instance. 🤷‍♂

PR #1852, opened 2020-03-25
Update UAT oEmbed provider.

jkeck:
It appears that it requires https (unlike the production provider)

### DIFF
--- a/config/initializers/oembed_providers.rb
+++ b/config/initializers/oembed_providers.rb
@@ -7,7 +7,8 @@ purl_provider << 'http://purl.stanford.edu/*'
 purl_provider << 'https://purl.stanford.edu/*'
 purl_provider << 'http://searchworks.stanford.edu/*'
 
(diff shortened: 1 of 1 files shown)


config/initializers/oembed_providers.rb
```diff
@@ -7,7 +7,8 @@
 purl_provider << 'https://purl.stanford.edu/*'
 purl_provider << 'http://searchworks.stanford.edu/*'
 
-purl_uat_provider = OEmbed::Provider.new('http://sul-purl-uat.stanford.edu/embed.{format}?&hide_title=true')
+purl_uat_provider = OEmbed::Provider.new('https://sul-purl-uat.stanford.edu/embed.{format}?&hide_title=true')
+purl_uat_provider << 'http://sul-purl-uat.stanford.edu/*'
```
jkeck: I don't think this addl. URL pattern is necessary since we don't pass `http` anywhere, but added here for consistency w/ the production URLs above.

PR #1857, opened 2020-03-30
Update dependencies

sul-devops-team:


jkeck:
These failures are probably due to sul-dlss/sul-embed#1131 and would be expected. The `src` of the iframe won't have that parameter anymore, and it will only be on the height attribute of the iframe.

PR #1866, opened 2020-04-08
Add full text indexing of hOCR content; fixes #1357

### LINKED ISSUE #1357: Support full text search/indexing of hOCR
This builds on #1039, and adds support for hOCR, the format used by Google Books. 

Assumptions:

* hOCR files will have `role="transcription"` associated with them in their contentMetadata

As an exhibit creator:

* When I add an SDR object with hOCR files associated with it, I want the text within those files to be searchable within the exhibit.

As a user:

* I want to be able to search the full text of a item, and find items that whose full text matches my search terms.

cbeer:


app/models/full_text_parser.rb
```diff
@@ -24,6 +24,8 @@ def to_text
       content = full_text_content(url)
       if ['application/xml', 'application/alto+xml'].include?(resource['mimetype'])
         alto_xml_string_content(content)
+      elsif ['text/html'].include?(resource['mimetype'])
```
jkeck, line 27: This isn't meant to hold this up (as I don't think we can do anything about it now), but certainly this will end up having us parse things that are not HOCR.  That shouldn't really matter in the grand scheme of things (as it likely won't have any `.ocr_page` content), but, thinking from the IIIF/Purl/Mirador perspective this way of determining HOCR might end up causing issues down the road WRT determining if we should advertise a search service for an object.

A quick spot check over argo indicates that this isn't an issue now, but not sure if we should flag this for thoughts WRT how we identify this flavor of OCR in the repository.

PR #1887, opened 2020-06-18
Add icon and styling for special exhibit filter

ggeisler:
Adds an icon and CSS styling for a new landing page filter called "Social justice":

Screen Shot 2020-06-18 at 2 50 49 PM

This, of course, requires us to add an exhibit tag labelled "Social justice" to applicable exhibits, which I think @caaster will do.


app/assets/stylesheets/modules/spotlight_overrides.scss
```diff
@@ -273,6 +273,22 @@ h5,
...
   padding: 0.33rem 0.66rem;
+
+  &[href="/?tag=Social+justice"] {
+    color: $cardinal-red;
+    font-weight: 700;
+
+    &::before {
+      // Licensed (required to change color) from:
+      // https://thenounproject.com/search/?q=justice%20scale&i=3316424
+      content: url(social-justice.svg);
```
cbeer, line 284:
```suggestion
      content: image_url('social-justice.svg');
```

I /think/ we need to use `image_url` to play nice with the asset pipeline?

ggeisler: Ah, okay, thanks. It worked locally so I didn't think about it. But I pushed up an amendment.

PR #1899, opened 2020-07-28
Stop storing all_search fields

cbeer:
AFAICT, the comment about highlighting came from upstream in searchworks. We've never used that field for highlighting in exhibits (and, in fact, override `hl.fl` to just those full tet fields)


corylown:
We are displaying hit highlights for OCR content such as in this sample exhibit: https://exhibits.stanford.edu/jarndyce-test/feature/search-across

Need to verify whether this is powered by the all_search field or not before we do this.

PR #1904, opened 2020-08-01
Update to rails 6

jkeck:
Not sure what to make of the failing spec that's remaining, but we can pick this up later.


mejackreed:
Included in #2007

PR #1944, opened 2020-10-19
Index IIIF + uploaded images with our custom large_image_url_ssm field

### LINKED ISSUE #1920: Uploaded items aren't displayed in results > slideshow view
In the Say Their Names exhibit all items are uploaded items. We noticed that in the [search results slideshow view](https://exhibits.stanford.edu/saytheirnames/catalog?utf8=✓&exhibit_id=saytheirnames&search_field=search&q=), none of the items images display, although they do in the list and gallery views:

Screen Shot 2020-08-21 at 3 15 32 PM

I checked Spotlight running locally, and uploaded items **do** display in the slideshow view. So it seems this might be an Exhibits-specific issue.

To double-check, I added three uploaded items to the existing SDR items in [my test exhibit](https://exhibits.stanford.edu/gary-testing/catalog?q=brady%27s&search_field=search&utf8=✓&view=list) (first three are uploaded, second three are SDR): 

Screen Shot 2020-08-21 at 3 12 37 PM

---

While the three uploaded items do display in other result and show page views, their images do not display in the results slideshow view:

Screen Shot 2020-08-21 at 3 12 52 PM

---

When I navigate to the fourth item, which is an SDR item, it does display, as do the other SDR items:

Screen Shot 2020-08-21 at 3 23 23 PM

cbeer:
This allows us to support the slideshow view ; fixes #1920

### DIFF
--- a/app/models/spotlight/resources/iiif_manifest.rb
+++ b/app/models/spotlight/resources/iiif_manifest.rb
@@ -13,7 +13,7 @@ module Spotlight
         return unless thumbnail_field
 
         image = manifest['thumbnail'] if manifest['thumbnail'].present?
-        image ||= manifest.dig('sequences', 0, 'canvases', 0, 'images', 0, 'resource')
+        image ||= manifest['sequences'].dig(0, 'canvases', 0, 'images', 0, 'resource')
 
         return unless image
 
--- a/app/models/spotlight/resources/iiif_manifest.rb
+++ b/app/models/spotlight/resources/iiif_manifest.rb
@@ -12,20 +12,19 @@ module Spotlight
       def add_thumbnail_url
         return unless thumbnail_field
 
-        if manifest['thumbnail'].present?
-          if manifest&.[]('thumbnail')&.[]('service')&.[]('profile')&.match? 'api/image'
-            solr_hash[:thumbnail_square_url_ssm] =
-              "#{manifest&.[]('thumbnail')&.[]('service')&.[]('@id')}/full/100,100/0/default.jpg"
-          end
-          solr_hash[thumbnail_field] = manifest['thumbnail']['@id']
-        else
-          first_image_id = manifest&.[]('sequences')&.[](0)&.[]('canvases')&.[](0)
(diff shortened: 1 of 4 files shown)


app/models/spotlight/resources/iiif_manifest.rb
```diff
@@ -12,20 +12,19 @@ class IiifManifest
...
-          solr_hash[thumbnail_field] = manifest['thumbnail']['@id']
-        else
-          first_image_id = manifest&.[]('sequences')&.[](0)&.[]('canvases')&.[](0)
-            &.[]('images')&.[](0)&.[]('resource')&.[]('service')&.[]('@id')
-          if first_image_id
-            solr_hash[:thumbnail_square_url_ssm] = "#{first_image_id}/full/100,100/0/default.jpg"
-            solr_hash[thumbnail_field] = "#{first_image_id}/full/!400,400/0/default.jpg"
-          end
+        image = manifest['thumbnail'] if manifest['thumbnail'].present?
+        image ||= manifest.dig('sequences', 0, 'canvases', 0, 'images', 0, 'resource')
```
mejackreed, line 16: I'm getting an `undefined method `dig' for #<IIIF::Presentation::Manifest:0x00007fc380ed3028>` when trying to index https://cdm15808.contentdm.oclc.org/iiif/info/p15808coll19/518/manifest.json . Maybe we need a better test here?

mejackreed (changes requested):
One minor comment about trying to index a manifest

app/models/spotlight/resources/iiif_manifest.rb
cbeer: Maybe so. I guess that's why we had the... interesting.. alternative.

mejackreed (changes requested):
Ok.. still seeing this?

```
TypeError (#<Class:#<IIIF::Presentation::Sequence:0x00007fcd7d858728>> does not have #dig method):
```

mejackreed (approved):
👍 thanks for making this change

PR #1964, opened 2020-12-02
Remove hard dependency on content metadata

cbeer:


mejackreed (approved):

caaster:
@mejackreed will this be in exhibits-prod following regular Monday morning updates on 7 Dec?

PR #1970, opened 2021-01-12
Updates per_page options to be multiples of 12

### LINKED ISSUE #1942: Change the search results per_page option values?
Because virtually all of our Spotlight at Stanford exhibits include gallery view in the search result options, is there any reason why we shouldn't change the search results per_page options to numbers that are evenly divisible by 3 and 4, which are the number of items we display in a row when in gallery view (3 for normal search results and 4 for browse category results where we don't display the facet sidebar)?

Doing so would eliminate the empty spaces we (almost) always get in gallery view at the end of a results page:

Screen Shot 2020-10-15 at 3 37 29 PM

For example, if we configured things like this:

```
# Maximum number of results to show per page
max_per_page: 96,
# Options for the user for number of results to show per page
per_page: [12, 24, 48, 96],
```

In gallery view the rows would always fill out (except possibly on the last result page, depending on the total number of result items), which would look much better.

For the the list and mosaic views I don't think the per_page number affects how the displayed result items look. For slideshow view we currently show rows of 6 items, which also doesn't work to fill out rows completely with our currently per_page options but would with the options proposed above, since they are all multiples of 6.

The only downside I see is that the per_page menu options are a bit harder to scan when they don't end in zero:

Screen Shot 2020-10-15 at 3 47 38 PM

But I suspect users don't access this menu all that often, and the arguably minor additional effort in scanning the options might be a worthwhile tradeoff for the improved page layout, especially since our exhibits often use gallery view as the default layout?

mejackreed:
Fixes #1942 

**NOTE** this will change all existing default per page options in existing Exhibits

## After
Screen Shot 2021-01-11 at 5 27 12 PM

## Before
Screen Shot 2021-01-11 at 5 26 26 PM

@caaster @ggeisler just want to mention the note above. Can you sign off on this before we ship?

### DIFF
--- a/app/controllers/catalog_controller.rb
+++ b/app/controllers/catalog_controller.rb
@@ -73,6 +73,11 @@ class CatalogController < ApplicationController
       'hl.mergeContiguous' => true
     }
 
+    # Maximum number of results to show per page
+    config.max_per_page = 96
+    # Options for the user for number of results to show per page
+    config.per_page = [12, 24, 48, 96]
+
     config.default_autocomplete_solr_params = Spotlight::Engine.config.default_autocomplete_params
 
     config.add_results_collection_tool(:sort_widget)
--- a/db/migrate/20210112001713_change_per_page.rb
+++ b/db/migrate/20210112001713_change_per_page.rb
@@ -1,0 +1,20 @@
+class ChangePerPage < ActiveRecord::Migration[5.2]
+  def change
+    Spotlight::Exhibit.find_each do |exhibit|
+      config = exhibit.blacklight_configuration
+      case config.default_per_page
+      when 10
+        config.default_per_page = 12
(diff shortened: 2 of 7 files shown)


ggeisler (approved):
Looks fine to me. We don't have any features where we store parameters that include the per-page value, do we (e.g. something like saved search/browse category)?

I think we should do this no matter what, just want to make sure we know if we'll affect anything that included the per-page parameter.

mejackreed:
It is possible that per_page could be saved (I think) but I don't think would be a common case that could happen.

ggeisler:
I guess it really doesn't matter, now that I think about it. Even if we did store the per_page parameter somewhere this change wouldn't break anything.

camillevilla (approved):

Issue #1976, opened 2021-01-20
Exhibit Admins are not able to select pages when using the pages widget


caaster:
This appears to be a bug and was reported by Elizabeth Ryan for the Conservation Services exhibit: https://exhibits.stanford.edu/conservation. Exhibit Admin Sarah Newton reports experiencing the same problem with the same exhibit.

I did a Zoom screen share with Beth & Sarah, and I was able to observe the following behavior when Beth shared her screen:
- Navigate to exhibit home page, and click edit
- Add the pages widget to the home page
- Type in the name of a feature page
- The feature page does not pop up as a selection, no matter which feature page name is entered

Please note:
- this was working just before holiday break, as Beth created the first row of feature pages you see on the home page using the page widget
- this worked for me today - I am the person who created the second row of feature pages using the feature page widget - **and it should be noted I am an exhibit super admin**
- Beth & Sarah were able to delete a page from the pages widget if desired, and also change the order of feature pages

caaster:
I just had Spotlight Service team member Andria Olson successfully validate the bug described above, using this exhibit: https://exhibits.stanford.edu/drmc-sandbox

cbeer:
Related to https://github.com/CanCanCommunity/cancancan/issues/677, maybe?

PR #1977, opened 2021-01-21
Pin cancancan

cbeer:
There is an upstream regression in handling STI classes (like pages) that makes things all screwy for pages.


mejackreed:
I'm wondering.. fixes #1976

PR #1991, opened 2021-02-17
Allow the ExhibitIndexer to commit immediately after adds/deletes so …

jkeck:
…that the autocomplete is as up-to-date as possible

### DIFF
--- a/app/models/exhibit_indexer.rb
+++ b/app/models/exhibit_indexer.rb
@@ -6,10 +6,16 @@
 # ExhibitIndexer.new(exhibit)
 # There is an #add method that will add/update the document in the index for that exhibit.
 # There is also a #delete method that will delete the document for that exhibit (by id).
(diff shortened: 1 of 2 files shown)


app/models/exhibit_indexer.rb
```diff
@@ -18,10 +24,12 @@ def to_solr
 
   def delete
     self.class.solr_connection.delete_by_id(document_id)
+    self.class.solr_connection.commit if commit
   end
 
   def add
     self.class.solr_connection.add(to_solr)
+    self.class.solr_connection.commit if commit
```
jkeck, line 32: I tried to do this using `commitWithin` passed through as an arg to `#add` but that didn't seem to work (and `#delete_by_id` doesn't have an analogous API to pass a commitWithin option).

These are running in background jobs so it shouldn't be too problematic, and the `commit` attribute allows use to set that to false if we want to script bulk exhibit indexing/deletion and commit after manually.

PR #2002, opened 2021-02-24
Use title_display instead of title_full_display for titles on the …

### LINKED ISSUE #1959: Remove extraneous period at end of item title on item detail page
Not sure when this creeped into things, but it appears that we are displaying an extraneous period at the end of item titles on the item detail page:

### Exhibits
https://exhibits.stanford.edu/cartosym/catalog/hy195vd0097

Screen Shot 2020-11-17 at 6 14 44 PM

### MODS
No period in metadata:

Screen Shot 2020-11-17 at 6 20 23 PM

### PURL page
No period showing on the item's PURL page:

Screen Shot 2020-11-17 at 6 15 20 PM

jkeck:
…record/show view

Closes #1959 

According to the stanford-mods code:

https://github.com/sul-dlss/stanford-mods/blob/3ffbfb58a8357de3ba4595206d0bd54c83dd5993/lib/stanford-mods/searchworks.rb#L189-L193

The title_display is just like title_full_display but w/o the additional punctuation (and is what is being used in SW)

## Before
mk267cj8576-before

## After
mk267cj8576-after

### DIFF
--- a/app/controllers/catalog_controller.rb
+++ b/app/controllers/catalog_controller.rb
@@ -111,7 +111,7 @@ class CatalogController < ApplicationController
     config.index.square_image_field = :thumbnail_square_url_ssm
     config.index.slideshow_field = :large_image_url_ssm
 
-    config.show.title_field = 'title_full_display'
+    config.show.title_field = 'title_display'
     config.show.oembed_field = :url_fulltext
     config.show.tile_source_field = :content_metadata_image_iiif_info_ssm
 
--- a/spec/features/metadata_display_spec.rb
+++ b/spec/features/metadata_display_spec.rb
@@ -15,19 +15,17 @@ RSpec.feature 'Metadata display' do
 
     it 'view metadata link links through to page', js: false do
       click_link 'More details »'
-      expect(page).to have_link 'Afrique Physique.', href: document_url
-      expect(page).to have_css 'h3', text: 'Afrique Physique.'
+      expect(page).to have_link 'Afrique Physique', href: document_url
+      expect(page).to have_css 'h3', text: 'Afrique Physique'
       expect(page).not_to have_css 'dt', text: 'Title:'
-      expect(page).not_to have_css 'dd', text: 'Afrique Physique.'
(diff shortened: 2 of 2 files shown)


ggeisler:
@jkeck Nice. I wonder if we could apply the same fix to the Title field in the metadata display? Searchworks and Purl don't repeat the title in the metadata they way we can (if the curator leaves Title checked in Config > Metadata) so there isn't a comparable situation there, but I'm not sure we need to show the `title_full_display` in the metadata. 

That said, I'm not certain what other than a punctuation mark at the end of a title might be different between `title_display` and `title_full_display`? But I kind of don't like the inconsistency between the title value in the page header and the title value in the metadata.

jkeck:
@ggeisler yeah, we can, although I wonder if we'll end up running into issues with existing exhibits who have this field and/or have configured this field (which might make this need a little more consideration)

cbeer:
If we definitely don't want to show `title_full_display` anywhere, maybe we can just alias it in the config (and... stop indexing it?):

```
config.add_index_field 'title_full_display', field: 'title_display', label: 'Title'
```

jkeck:
Heh, I was just checking if adding `field` explicitly would work.  But yeah, agreed, that does seem to be the case that nobody wants to use the `title_full_display` field (although I think a little deeper analysis would be warranted)

ggeisler:
I feel fairly certain that if we changed the value of title in the metadata display to `title_display` no existing curator would notice or care (I think almost always the curator should hide the title from the metadata display since it's already prominent on the page as the page title, but most just neglect to turn it off). So from that point of view I'm good with switching it. But if there are technical or other considerations that curators might care about that I'm not realizing are in play here, let me know.

jkeck:
The concern I was thinking mostly of was technical, but I realized after I typed it (and as @cbeer notes) there is a `field` key that we could pass to our existing configured `title_full_display` field to more-or-less "alias" it to our `title_display` field. 

There is some other code that uses the notion of the `title_full_display` field, and just sorting through some of that now.

PR #2004, opened 2021-02-25
Use the title_display field in place of the title_full_display field

jkeck:
The title_full_display has punctuation that we don't want to display

We are keeping the configured key as title_full_display so that nothing
changes for current exhibits but they will now use the title_display
field from the solr document


jkeck:
Note that https://github.com/sul-dlss/exhibits/blob/7f16fcf66650147c4137f745268994c4b5f5ed0e/app/assets/javascripts/cited_documents.js#L22 is using the Blacklight 7-style JSON API which uses the configured key. Since we're keeping that configuration key, code that uses the API will need to continue to use that key.

PR #2015, opened 2021-03-02
Remove SidekiqValidityChecker as we now have better JobTracking upstr…

mejackreed:
…eam and this was causing things to have problems


cbeer:
🤷‍♂️

PR #2018, opened 2021-03-03
Removes Mirador 2 viewer fixes #1990

### LINKED ISSUE #1990: Deprecate Mirador 2 viewer
Due to security issues with the current version of Mirador 2 running in Exhibits we will need to deprecate the ability for users to use this viewer in their exhibit.

Current exhibits (as of 2021-02-12) that use the viewer: ["parker", "time-travel-japan", "burke_mss", "secret-service", "ars-test", "epl", "medieval", "service-sandbox", "tress", "paleography", "test-lgbtqdhh"]

Likely steps that need to be taken:
 - Contacting exhibit curators
 - Disabling the ability to select Mirador 2 in curator viewer config. https://github.com/sul-dlss/exhibits/blob/master/app/views/viewers/edit.html.erb#L27
 - Migrate exhibits to using Mirador 3.

Mirador 3 is already an option that can be chosen, and curators have the ability to make this change on their own.

mejackreed:
This pull request removes the legacy Mirador 2 viewer. All Exhibits have been migrated, and there is a fallback just in case to render the mirador3 viewer instead.


app/models/viewer.rb
```diff
@@ -17,6 +17,8 @@ def default_viewer_path
   def to_partial_path
     return default_viewer_path if viewer_type == 'sul-embed' || viewer_type.nil?
 
-    "../viewers/#{viewer_type}"
+    modified_viewer_type = viewer_type == 'mirador' ? 'mirador3' : viewer_type
```
mejackreed, line 20: We could fallback a different way.. open to suggestions here.

Gemfile
```diff
@@ -125,7 +125,6 @@ gem 'stanford-mods', '~> 2.2', '>= 2.2.1'
 gem 'solrizer'
 gem 'honeybadger'
 gem 'slowpoke'
-gem 'mirador_rails'
```
camillevilla, line 128: 👏

app/models/viewer.rb
camillevilla: This seems... reasonable enough for now? Could we throw a data remediation on the pile for our next maintenance work cycle?

camillevilla (approved):

app/models/viewer.rb
mejackreed: Forgot to mention, I just checked in prod (about an hour ago) and no more Exhibits had this.. that is why I just went with this.

camillevilla:
@caaster I wanted to flag this for you. I don't think curators are choosing viewers without the guidance of the service team, but the service team may need to update the language and screenshots in the [Exhibits documentation for Configuration > Viewers](https://exhibits.stanford.edu/exhibits-documentation/feature/viewers). 

To re-cap where things stand at the moment, since both viewers are technically Mirador 3 with a Stanford color theme... 
SUL-embed
- only option appropriate for audio / video, files, and PDF content
Mirador 
- includes image tools (rotation, flip, contrast, brightness, saturation, greyscale, invert colors)

### Before
Screen Shot 2021-03-03 at 4 02 45 PM

### After
Screen Shot 2021-03-03 at 4 16 39 PM

PR #2031, opened 2021-03-08
Hacky attempt to enable the map in BrowseController fixes #1876

### LINKED ISSUE #1876: Map view doesn't display geolocated items for Browse categories
I thought this was fixed in the Spotlight workcycle earlier this year, but it appears to be broken again.

See for example: https://exhibits.stanford.edu/lanciani/browse/roma-xi-1-colosseo-f7f5a2f1-9b7e-459e-b08a-431ad37cff96?view=heatmaps

mejackreed:
There is probably a better way here

### DIFF
--- a/app/helpers/blacklight/maps_helper_override.rb
+++ b/app/helpers/blacklight/maps_helper_override.rb
@@ -13,5 +13,15 @@ module Blacklight
       original_path = spotlight.exhibit_solr_document_path(exhibit_id: current_exhibit, id: token)
       original_path.sub(token, "{#{blacklight_config.document_model.unique_key}}")
     end
+
+    def index_map_data_attributes
+      attributes = super
+      if controller.is_a? Spotlight::BrowseController
+        attributes[:search_url] = spotlight.search_exhibit_catalog_url(
+          controller.send(:search_query).merge(search_field: 'search')
+        )
+      end
+      attributes
+    end
   end
 end
--- a/spec/features/heatmaps_spec.rb
+++ b/spec/features/heatmaps_spec.rb
@@ -1,0 +1,26 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.feature 'Blacklight heatmaps', js: true do
+  let!(:exhibit) { create(:exhibit, slug: 'default-exhibit') }
+
+  describe 'accessing from CatalogController context' do
+    it 'renders the leaflet map with results' do
+      visit spotlight.search_exhibit_catalog_path(exhibit, q: ' ', view: 'heatmaps', search_field: 'all_fields')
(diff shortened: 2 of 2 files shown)


app/helpers/blacklight/maps_helper_override.rb
```diff
@@ -13,5 +13,15 @@ def document_path
       original_path = spotlight.exhibit_solr_document_path(exhibit_id: current_exhibit, id: token)
       original_path.sub(token, "{#{blacklight_config.document_model.unique_key}}")
     end
+
+    def index_map_data_attributes
```
cbeer, line 17: I don't suppose we could get an integration test for this?

And, let me see if I've got this right: for browse categories, we're redirecting the `search_url` parameter over to the catalog controller with the search parameters of this browse category,  because... the browse category url doesn't respond to the type of requests heatmaps is making (and... uh.. what requests are those?)

I might have a better suggestion based on the type of request, but I also wonder if we want to pass the browse category over to the search_exhibit_catalog_url and de-reference it there? It might be useful for bulk-actions-on-browse-categories, 🤷

cbeer (changes requested):

app/helpers/blacklight/maps_helper_override.rb
mejackreed: I've pushed up a commit with some tests around the integration piece.

> because... the browse category url doesn't respond to the type of requests heatmaps is making

Yep that seems part of it.. and also because naively in heatmaps we just use `request.url` to figure out what to do things on.

https://github.com/sul-dlss/blacklight_heatmaps/blob/master/app/helpers/blacklight/maps_helper.rb#L45

I'm not sure how that suggestion would work, but do we need to do something better upstream?

cbeer: The missing piece for me is... what do we do with `search_url`?

mejackreed: We just pop on a `view=heatmaps` and clientside request away: https://github.com/sul-dlss/blacklight_heatmaps/blob/master/app/assets/javascripts/blacklight_heatmaps/viewers/index.js#L18

cbeer: Got it (searching for search_url didn't turn that up..). And browse categories don't response to `format=heatmaps`?

mejackreed: Yeah.. getting a 

Screen Shot 2021-03-10 at 11 42 32 AM

cbeer: Ah, no, that's my copying the unreleased feature over instead of the 7.15-compatible version.

PR #2032, opened 2021-03-08
Use purl based display logic for Mods display

### LINKED ISSUE #1051: Display each instance of abstract element on new line instead of concatenating
Abstracts are often a prose paragraph of complete sentences ending with a period, so concatenation with "and" is not appropriate. Per LC MODS guidelines, this field is repeatable, and in particular MODS records converted from MARC may have multiple instances of the abstract field.

Example: https://exhibits.stanford.edu/time-travel-japan/catalog/fr878dp2426

Related issue in mods_display gem: https://github.com/sul-dlss/mods_display/issues/54

### LINKED ISSUE #1958: Update how multiple Note elements are rendered to HTML in MODS display
(I'm not sure if this is an Exhibits or a MODS-display issue; I can move it to another repo if a developer wants me to.)

In the item detail page, when selecting the More Details metadata button to see the full metadata, if there are multiple `<note displayLabel="Note">` elements they are correctly grouped under a "Note" heading, but there are no line breaks (or vertical spacing to clearly separate) between the individual element values. This makes the text more difficult to read and doesn't reflect the cataloger's intent to make these distinct notes.

### Source
https://purl.stanford.edu/rj493vx4753.mods

Screen Shot 2020-11-17 at 5 47 42 PM

### Spotlight rendering
https://exhibits.stanford.edu/james-e-allen/catalog/rj493vx4753

Screen Shot 2020-11-17 at 5 44 05 PM

### PURL page rendering
https://purl.stanford.edu/rj493vx4753

Ideally multiple note elements would be rendered similarly to how they are in PURL. Here's the PURL page for the same object as above, where there is vertical spacing between each note:

Screen Shot 2020-11-17 at 5 45 05 PM

It looks like the difference is due to how multiple note elements are transformed to HTML in the two cases:

- PURL: each note element is wrapped **in its own** `dd` element, all siblings to the same "Note" title `dt` element
- Spotlight: each note element is included in a **single** `dd` element that is a sibling to the same "Note" title `dt` element

The PURL approach of using multiple `<dd>`s seems more ideal to me, but I don't know what technical constraints might be involved.

### LINKED ISSUE #1961: Support line break characters in metadata field values
Sort of related to #1958. This issue is visible in Exhibits, but is also an issue in Searchworks so the solution might not be specific to Exhibits but I'm filing here because I'm not sure where the most appropriate place would be.

When line break characters are in the MODS, they are respected in the rendered output on the PURL page.

Those characters are not respected, however, in Spotlight or Searchworks. We know that in at least one upcoming exhibit the curator has good reason for wanting line breaks to be respected. Plus it seems like it would be good to be consistent across applications.

In the example below, the line break characters are used in the `<abstract>` element, but they are likely to be used in at least the upcoming exhibit in the `<note>` element as well, so ideally the request to respect line breaks is global, rather than field specific. 

Example used here is: https://purl.stanford.edu/bc777tp9978

### MODS with line break characters
Screen Shot 2020-11-20 at 10 42 59 AM

### PURL page showing line breaks
Screen Shot 2020-11-20 at 10 40 35 AM

### Spotlight detail page, without line breaks
Screen Shot 2020-11-20 at 10 41 14 AM

### Spotlight "more details" full metadata modal
Screen Shot 2020-11-20 at 10 41 47 AM

### Searchworks detail page
Screen Shot 2020-11-20 at 10 44 15 AM

mejackreed:
Fixes #1961 Fixes #1051 Fixes #1958

Requires https://github.com/sul-dlss/mods_display/pull/73

This pull request brings inline the Exhibits feature of "More details" to show the mods record rendered as HTML to display as it does in the purl application.

It solves a number of issues from over the years, where Exhibit curators expect this metadata to render as it does on the Purl page.

However, the display is changed from what was previously there and before moving this forward I want to make sure there is awareness of the impact.

Another thing this does is for "Description" show/index views, they are now joined by `<p>` and new lines within the metadata are split also.

Examples of how updates look:

### wz026zp2442 Parker example with nested items
Current: https://parker.stanford.edu/parker/catalog/wz026zp2442/metadata
Update: https://exhibits-stage.stanford.edu/mods-display-test-exhibit/catalog/wz026zp2442/metadata

### bc777tp9978 abstract with line breaks #1961
Current: Not currently in an exhibit
Update: https://exhibits-stage.stanford.edu/mods-display-test-exhibit/catalog/bc777tp9978/metadata

### fr878dp2426 abstract elements should be on new lines #1051
Current: https://exhibits.stanford.edu/time-travel-japan/catalog/fr878dp2426/metadata
Update: https://exhibits-stage.stanford.edu/mods-display-test-exhibit/catalog/fr878dp2426/metadata

### rj493vx4753 multiple note elements as paragraphs #1958
Current: https://exhibits.stanford.edu/james-e-allen/catalog/rj493vx4753/metadata
Update: https://exhibits-stage.stanford.edu/mods-display-test-exhibit/catalog/rj493vx4753/metadata

Because of the potential impact on Parker, I wanted to give @blalbrit a heads up in advance of anything moving forward.


ggeisler:
@mejackreed Based on the examples you provided, the updates look great to me.

Regarding @blalbrit's comment:

> the only request I would have would be to use the displayLabel in the Abstract/Summary and Table of Contents elements. I don’t want to presume, but if displayLabel exists for those fields, maybe we could use it and then default to Abstract/Contents if it’s not present?

This sounds reasonable to me. It would reduce consistency of the Abstract/Summary and Table of Contents headings across objects but presumably the cataloger used displayLabel for a reason and if their displayLabel heading makes more sense for that object's values it seems reasonable to me to use it.

This would affect Purl as well, I presume, since our goal is to make the display consistent across applications.

I have a separate question after looking at the comparisons that it is specific to #1961. The example for this is  https://exhibits-stage.stanford.edu/mods-display-test-exhibit/catalog/bc777tp9978/metadata. That example, which uses the abstract field to show the improvement, is exactly what we want. But in the ticket description of #1961 I noted:

> In the example below, the line break characters are used in the `<abstract>` element, but they are likely to be used in at least the upcoming exhibit in the `<note>` element as well, so ideally the request to respect line breaks is global, rather than field specific.

So I'm wondering, does the PR only solve this for the `<abstract>` element or will it also be applied to `<note>` elements that include line breaks?

mejackreed:
@ggeisler do you know of any content that has `<note>` like that? Yes it should display like that correctly, but if we have an example I would be happy to confirm.

Here is what it would look like showing the display label:

Screen Shot 2021-03-09 at 2 56 46 PM

ggeisler:
@mejackreed The display label adjustment looks great to me. I don't see any reason not to do that.

For the `<note>` issue, no, I don't know of an example object that currently has line-break characters within a `<note>` element but I know the Wong project wants to do it. They are basically waiting for us to say we will support it and then they will start using line-breaks in their note fields (they are in the middle of metadata preparation for 600 or so items, I believe).

As an example of what they want to do, and what Lynn has done in an existing record that would be improved by supporting line breaks in the `<note>` element, see this record: https://exhibits-stage.stanford.edu/mods-display-test-exhibit/catalog/rj493vx4753/metadata

See the `<note>` element with the displayLabel of "Exhibits" in particular. I believe Lynn embedded the character for a bullet to separate pieces of the value that he would've preferred to be on separate lines. So if you replace the bullet characters with the line-break characters in that field, that would be a realistic example.

In the Wong project they would be doing the exact same thing. They want to list multiple art installations or exhibits in a single note field but display them on separate lines to make them easier for a user to parse.

mejackreed:
@ggeisler I've updated our stage environment with the display labels

ggeisler:
👍 Looks good to me.

caaster:
Hi maybe i am off base here, but i had understood the request as not only for multiple notes etc. to appear on a new line, but for there to be a space in between each entry in addition. Please reality check me here.

mejackreed:
@caaster they are now displayed consistently with Purl. See the screenshot. If you are looking at a deployed environment at the moment it may not represent this code.

Screen Shot 2021-03-11 at 9 05 01 AM

jkeck (approved):
👍

PR #2038, opened 2021-03-11
Remove unnecessary CSS selector

ggeisler:
There is a noticeable shifting in size of the exhibit title text in the masthead when you move from to and from a Browse page from any other page:

https://user-images.githubusercontent.com/101482/110867883-c2c64600-8284-11eb-86a1-9013aa3f1884.mov

I thought I had a fix but realized I need to look at one other potential issue involved so am making this a draft for now.

### DIFF
--- a/app/assets/stylesheets/modules/spotlight_overrides.scss
+++ b/app/assets/stylesheets/modules/spotlight_overrides.scss
@@ -97,10 +97,6 @@
   }
 }
 
-.blacklight-browse h1 {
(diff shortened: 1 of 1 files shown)


ggeisler:
I'm going to close this PR because there's something weird going on with my local dev environment for Exhibits that makes it hard for me to be 100% sure this is a complete fix for the problem.

I'll create an issue for the problem instead and maybe someone else can take a look at it.

Issue #2052, opened 2021-03-25
Masonry view has lost its mosaic-arrangement visual qualities


caaster:
Please see some examples:

- https://exhibits.stanford.edu/maps/browse/united-states?view=masonry
- https://exhibits.stanford.edu/herbert-matter/browse/all-exhibit-items?page=8&view=masonry
- https://exhibits.stanford.edu/fitch/browse/martin-luther-king-jr-1965-1966?view=masonry

camillevilla:
FYI there has been progress on this but it required some upstream work in projectblacklight/blacklight-gallery#99

mejackreed:
@cbeer @camillevilla are we good to close this now?

cbeer:
Looks mosaic-y to me 🤷‍♂️

PR #2063, opened 2021-04-01
Remove ancient thumbnail override

cbeer:
This was initially added for #1116, but now breaks #2052.. Since then, I think we've picked square thumbnails that are 100x100 anyway, so I can't see it being useful still 🤷‍♂️

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -95,7 +95,7 @@ gem 'config'
 
 gem 'bootstrap'
 gem 'blacklight', '~> 7.15'
-gem 'blacklight-gallery', '~> 3'
+gem 'blacklight-gallery', '~> 3', '>= 3.0.3'
 gem 'blacklight_heatmaps'
 gem 'blacklight-spotlight', '~> 3.0.0.rc3', github: 'projectblacklight/spotlight'
 gem 'twitter-typeahead-rails', '0.11.1.pre.corejavascript'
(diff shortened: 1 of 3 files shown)


mejackreed:
Just pushed another commit.. now things look like:

Screen Shot 2021-04-01 at 1 49 14 PM

ggeisler:
Could we reduce the size of the gutters between columns? Generally mosaic views show the images very tightly packed, with only a small (~6px?) gutter between columns.

I'm also wondering why the image in the second column, third row doesn't take full-width of the column. Unless the whitespace on the right edge of that image is part of the image background, I think the desired behavior is for it to take full-width, even if that results in it being taller. Again, the idea is to fill the space with image content and not show much visible whitespace.

cbeer:
Without measuring, I'd guess that's the full size of the thumbnail -- we're just requesting 400-pixel-long-edge images

cbeer:
(I'd also guess this is pre-existing behavior and not part of addressing the regression from updating blacklight-gallery to use view components... at least, I don't see where we've ever indexed a fixed-width version of the thumbnail 🤷‍♂️ )

ggeisler:
I see, that makes sense.

The wider gutters are new, though, so hopefully we can fix those?

mejackreed (approved):

PR #2085, opened 2021-05-03
Avoid inadvertently upscaling deliberately small images

### LINKED ISSUE #2077: Prevent uploaded images from being upscaled
@caaster and I noticed that when uploading images (using the Uploaded item row widget) that are intentionally small (width ~300px), the rendered version of the image is much larger (2000px on the long side). I assume this has to do with the `resize_to_fit` method in the [configuration](config/application.rb): `Spotlight::AttachmentUploader.send(:process, resize_to_fit: [2000, 2000])`.

Ideally, we want to stick with the decision to resize uploaded images to 2000px on the long side, but only when that is _downscaling_ to prevent overly large images. So the optimal solution is to use `resize_to_fit` when the uploaded image comes in at greater than 2000px on the long side, but to omit the `resize_to_fit` method (or some equivalent strategy) when the uploaded image is smaller than 2000px on its long side.

Rationale for this: 

- sometimes the curator will want to display an uploaded image that is intentionally small (it's an image detail or otherwise a supplemental image not appropriate for a large rendered size)
- upscaled images lose image quality and can look ugly

Example of problem, where the second image is from pre-resize_to_fit method, and is the size we want to display, while the first image was the same uploaded size but now displays larger than desired (and it is really big if we display it on its own in the row, without the text next to it):

Screen Shot 2021-04-13 at 3 38 41 PM

mjgiarlo:
Fixes #2077

Use the CarrierWave `resize_to_limit` method instead of `resize_to_fit`. The difference? The latter upscales and the former does not.

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -25,7 +25,7 @@ module Exhibits
     config.to_prepare do
       Spotlight::Exhibit.send(:include, ExhibitExtension)
       Spotlight::AttachmentUploader.send(:include, CarrierWave::MiniMagick)
-      Spotlight::AttachmentUploader.send(:process, resize_to_fit: [2000, 2000])
+      Spotlight::AttachmentUploader.send(:process, resize_to_limit: [2000, 2000])
     end
 
     config.druid_regex = /([a-z]{2}[0-9]{3}[a-z]{2}[0-9]{4})/


ggeisler:
@mjgiarlo Yep, this seems to do what we want. Below is a source image of 300x468 dimensions, which is rendered at its native size on -stage. Previously, if this was uploaded to a Spotlight feature page using the Upload item row widget, it would be scaled up to render at size much larger than the native image size:

Screen Shot 2021-05-03 at 4 12 38 PM

---

I also uploaded an image greater than 2000px on the long side and it was correctly downscaled to 2000px, so it also appears your fix didn't break anything when the image exceeds 2000px.

So all looks good to me!

mjgiarlo:
@ggeisler much obliged, sir.

Issue #2138, opened 2021-10-28
Exhibit pages unresponsive to editing and freeze using Google Chrome


caaster:
I am corresponding with an exhibit creator in Estonia, who has filed a feedback ticket regarding their experience with the [Vabamu exhibit](https://exhibits.stanford.edu/vabamu). Please see the email thread below my comments. I feel like I could use more information (the time zone difference and difficulty of setting up a Zoom session pose quite the delay, however), so I started investigating this morning.

This is my experience, below. Using Mac OS Catalina, 10.15.7, Google Chrome Version 95.0.4638.54 (Official Build) (x86_64).

- I navigate to this exhibit: [Cathy's test 2](https://exhibits.stanford.edu/cathy-s-test-2)

- I click on "Edit"

- I attempt to click on the plus sign next to the widget box to display the widget options. My attempts to click are unresponsive.

- Eventually, the following message appears:

![Screen Shot 2021-10-28 at 8 43 27 AM](https://user-images.githubusercontent.com/3210225/139291481-3b75679a-0fe0-433f-b9a0-79027c9bbb7f.png)

I am able to repeat this behavior on other exhibits using Chrome, but I am not able to repeat the behavior using Safari Version 15.0 (15612.1.29.41.4, 15612).

- What I described above seems a likely similar behavior to the experience Karl-Erik describes below.

*************

From: Karl-Erik Laurents <laurents@tlu.ee>
Sent: Thursday, October 28, 2021 6:14 AM
To: Catherine A. Aster <caster@stanford.edu>
Subject: Re: [exhibits-feedback] VABAMU exhibit feedback
 
Hello!

1. The problem still exists 
2. I am using Windows 10 Home and Google Chrome Version 95.0.4638.54 (Official Build) (64-bit)
3. I was trying to place exhibit picture, when the first crash occured.
4. The crash occures whenever i am clicking on edit (trough home page or trough exihibt dashboard) it takes me to edit page and the whole site crashes and chrome notifies me that the page is unresponsive. I am adding the picture of Chromes notification below. The crash only happens on edit page. 
image.png

Kontakt Catherine A. Aster (<caster@stanford.edu>) kirjutas kuupäeval K, 27. oktoober 2021 kell 22:45:
Hi Karl-Erik,

Thank you for reporting the problem with the Vabamu exhibit. 

I have a few questions:
Are you still experiencing the problem you described?
What Operating System and Browser are you using? For example, this is the level of detail we're asking for:
[...]

PR #2139, opened 2021-10-28
Override SirTrevor to work around a niche chrome-specific bug

cbeer:
If using another browser isn't possible, this is a maybe-not-a-good-idea way to fix the behavior in chrome.

See #2138.


camillevilla:
Deploying this to stage for some poking

camillevilla (approved):

camillevilla:
@caaster This should allow curators to use Chrome again. This override targets Chrome users with versions 95-96.

camillevilla:
🚢 This has been deployed to prod, @caaster.

PR #125, opened 2015-11-30
add new document subtype field to facet and display

peetucket:

PR #2159, opened 2022-01-27
Push indexing logic downstream where exhibits is the only data consumer

cbeer:


thatbudakguy (approved):
this is kind of hard to interpret without comments or an explanation in the PR, but I'm assuming:

- exhibits does some special formatting of these date/title/etc values to make them look good in various facets or display nicely for various collections
- the logic that did this used to be "upstream" in one of our mods gems, but since exhibits was the only consumer, we removed it from there (not shown in this PR) and moved it to here
- rather than sticking it in a random file, we opted to just have it live directly in the traject config

if all these are true then I think I understand this enough to merge.

camillevilla (approved):
Good summary for @thatbudakguy. Here's the relevant commit from upstream sul-dlss/stanford-mods for reference.
[sul-dlss/stanford-mods@`b053e08` (#125)](https://github.com/sul-dlss/stanford-mods/pull/125/commits/b053e08675f3c5d8934d0c33a8bcb64208e919d1)

PR #2195, opened 2022-06-13
Update dependencies

sul-devops-team:


yarn.lock
```diff
@@ -0,0 +1,4 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
+# yarn lockfile v1
+
+
```
thatbudakguy, line 4: empty yarn.lock?

thatbudakguy: oh, is this because it warns when there's no lockfile?

corylown (approved):

PR #2273, opened 2023-01-19
Load rails 7 defaults

jcoyne:


spec/controllers/login_controller_spec.rb
```diff
@@ -20,12 +20,12 @@
 
   context 'with a referrer header' do
     before do
-      request.env['HTTP_REFERER'] = 'http://example.com'
+      request.env['HTTP_REFERER'] = 'http://test.host/foo'
```
jcoyne, line 23: Rails 7 not protects us from redirecting to other sites. See https://blog.saeloun.com/2022/02/08/rails-7-raise-unsafe-redirect-error.html

corylown (approved):

PR #2283, opened 2023-01-24
Inline sul_styles

jcoyne:
This helps us so that we can migrate from libsass to dart sass


corylown:
Work completed via https://github.com/sul-dlss/exhibits/pull/2663 which removed sul-styles.

PR #2316, opened 2023-05-12
Test with Ruby 3.2

cbeer:


jcoyne:
@cbeer I believe the problem in the tests was caused by code we have in this repo:
https://github.com/sul-dlss/exhibits/blob/636e1eed9da3f976d58309c8f35916d9bffbfeac/config/settings.yml#L69

PR #2322, opened 2023-06-02
Update nav link to information about Spotlight at Stanford. Closes #2321

### LINKED ISSUE #2321: Change link in About dropdown on exhibits landing page
**Requested action:** On the https://exhibits.stanford.edu/ landing page, under the About dropdown menu, please delete the link to theSpotlight at Stanford microsite. In its place, please put the following link text + link: How does Stanford use Spotlight? https://exhibits.stanford.edu/exhibits-documentation/about/how-does-stanford-use-spotlight

Screen Shot 2023-05-30 at 11 21 40 AM

**Reason for request:** The Spotlight microsite will no longer be supported when the new library website launches on 31 July 2023. All microsite content has been moved to this exhibit: https://exhibits.stanford.edu/exhibits-documentation. The highest value content on the microsite is "How does Stanford use Spotlight?"

**Date the change is needed by:** No later than 27 July 2023. It would be completely ok for this change to happen now, as well.

corylown:


app/views/shared/_site_sidebar.html.erb
```diff
@@ -10,8 +10,8 @@
 
   <h2 class="nav-heading">See also</h2>
   <ul class="nav sidenav flex-column">
```
jcoyne, line 12: Can we remove this page entirely?  It seems to be part of a feature flag (https://github.com/sul-dlss/exhibits/blob/6fbae9c200cc152b0f7bca72043fd2dc48c2713c/app/views/shared/_site_sidebar.html.erb#LL1C1-L1C1), but I don't think we're ever going to flip it back.

jcoyne (approved):

app/views/shared/_site_sidebar.html.erb
corylown: I'm not sure, but I'll create an issue to track it.

corylown: https://github.com/sul-dlss/exhibits/issues/2323

PR #2335, opened 2023-07-28
Remove home_page_navbar feature flag

jcoyne:
It's been in production for 3 years


app/views/shared/_site_navbar.html.erb
```diff
@@ -1,76 +1,74 @@
-<% if Settings.feature_flags.home_page_navbar %>
```
jcoyne, line 1: This is easier to read when using "Hide whitespace"

thatbudakguy (approved):

PR #2394, opened 2024-02-05
Update dependencies

sul-devops-team:


Gemfile
```diff
@@ -91,7 +91,7 @@ gem 'bootstrap', '~> 4.6'
 gem 'blacklight', '~> 7.34'
 gem 'blacklight-gallery', '~> 4.4'
 gem 'blacklight_heatmaps'
-gem 'blacklight-spotlight', '~> 3.0'
+gem 'blacklight-spotlight', '~> 3.5.0' # Stay on spotlight 3.5.* until we do the work to update to 3.6
```
jcoyne, line 94: What work is required?  There shouldn't be anything necessary for a minor release.  If there are some breaking changes, we should consider yanking this release and releasing as a major version.

corylown: 3.0.6.beta7 doesn't just work without any changes to exhibits. I'm a bit confused about why bundler updated from `3.5.0.3` to a beta release of `3.6.0.beta7`. I thought the `~> ` notation would exclude a pre/beta release unless a pre-release was explicitly specified.

cbeer: Breaking changes in `3.6` seems like a problem that needs to be fixed upstream.

This seems like a good approach to stopping our stuff from breaking, but it seems like something ought to be reported upstream?

PR #2407, opened 2024-04-08
Update IIIF link to https://iiif.io

corylown:
After some discussion with Sarah we decided to update the IIIF links to https://iiif.io

### DIFF
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -31,7 +31,7 @@ traject:
   processing_thread_pool: 1
 iiif_embed:
   url: https://embed.stanford.edu/iiif
-iiif_dnd_base_url: https://library.stanford.edu/projects/international-image-interoperability-framework/viewers?%{query}
+iiif_dnd_base_url: https://iiif.io?%{query}
 action_mailer:
   default_options:
     from: "example-team@example.com"


config/settings.yml
```diff
@@ -31,7 +31,7 @@ traject:
   processing_thread_pool: 1
 iiif_embed:
   url: https://embed.stanford.edu/iiif
-iiif_dnd_base_url: https://library.stanford.edu/projects/international-image-interoperability-framework/viewers?%{query}
+iiif_dnd_base_url: https://iiif.io?%{query}
```
jcoyne, line 34: Should we remove the `query` stuff. This page isn't going to support manifests like the old one did.

jcoyne (changes requested):

config/settings.yml
corylown: I don't think the previous page we linked to did anything with the query parameter either, but it's needed so that the drag and drop feature will work when there's a manifest referenced in the query parameter, such as:

`https://iiif.io?manifest=https%3A%2F%2Fdms-data.stanford.edu%2Fdata%2Fmanifests%2FParker%2Fqd527zm3425%2Fmanifest.json`

jcoyne: can we remove all of this then? https://github.com/sul-dlss/exhibits/blob/b1f6d58a1ce12f1f78f9f4844d34558f1fdb76f8/app/helpers/application_helper.rb#L22

thatbudakguy: My understanding of this is that we have to make a full link to make drag&drop work, but the host/path part of the URL is irrelevant because the user isn't supposed to actually _click_ the link, only drag it. If they do, we may as well send them to a page that is at least related to IIIF. That's what the old page did – it didn't use the manifest in any way – and that URL now redirects to the new page anyway. All this change does is make the redirect explicit instead of implicit, right?

corylown: Original issue explaining the design of the IIIF icon: https://github.com/sul-dlss/exhibits/issues/679

As Nick said, it's designed so that when someone clicks on the icon it goes to a page explaining with IIIF is and if the user drags the icon onto a viewer the manifest provided in the parameter adds the resource to the viewer.

PR #2414, opened 2024-04-19
Remove unused delegate

jcoyne:


jcoyne:
This is used for the `imprint` in dor_config (traject)

PR #2416, opened 2024-04-19
Make the PublicXmlRecord local

jcoyne:
This will be removed from purl-fetcher in a future release.


corylown:
@jcoyne is this something you still want to implement in exhibits?

jcoyne:
@corylown yes it is required because this class was removed from purl_fetcher-client in the 1.0.0 release:
 https://github.com/sul-dlss/purl_fetcher-client/commit/0d8d3c627bd8446ccb9a3fb6683617ff694314cf

corylown (approved):

PR #2421, opened 2024-05-08
Remove unnecessary indexing

jcoyne:
We now take care of all of this with a IIIF manifest. All public XML files already have the `thumb` node, so we don't need to duplicate this logic anymore.

### DIFF
--- a/lib/traject/dor_config.rb
+++ b/lib/traject/dor_config.rb
@@ -202,40 +202,6 @@ end
 
 to_field 'content_metadata_type_ssm', copy('content_metadata_type_ssim')
 
-each_record do |resource, context|
-  content_metadata = resource.public_xml.at_xpath('/publicObject/contentMetadata')
-  next if content_metadata.blank?
-
-  # Select conventional file images or virtual external ones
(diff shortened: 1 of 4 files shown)


lib/traject/dor_config.rb
```diff
@@ -202,38 +202,9 @@
...
-  accumulator << image_data['width']
-end
-
-to_field 'content_metadata_first_image_height_ssm' do |_resource, accumulator, context|
-  next unless context.clipboard['thumbnail_data']
-
-  image_data = context.clipboard['thumbnail_data'].at_xpath('./imageData')
-
-  accumulator << image_data['height']
 end
```
corylown, line 208: Did you mean to leave lines 205-208? I don't think they do anything now that the rest of that `each_record` block is removed.

corylown (changes requested):

lib/traject/dor_config.rb
jcoyne: Thanks. I've removed that now.

corylown (approved):

PR #2426, opened 2024-06-03
Update dependencies

sul-devops-team:


marlo-longley:
This test failure with the `newrealic_rpm` gem is a known issue with a submitted PR to fix it:
https://github.com/newrelic/newrelic-ruby-agent/issues/2675

I think we should wait for that to be merged, then try again to do dependency updates.

marlo-longley (approved):

PR #2559, opened 2024-09-20
switch to blacklight startover button component

### LINKED ISSUE #2557: start_over_path is deprecated
```
DEPRECATION WARNING: start_over_path is deprecated and will be removed from a future release (Removed without replacement). (called from _app_views_catalog__start_over_html_erb___1659957843188787546_126400 at /opt/app/exhibits/exhibits/releases/20240916110307/app/views/catalog/_start_over.html.erb:1)
```

dnoneill:
closes #2557

### DIFF
--- a/app/components/start_over_button_component.rb
+++ b/app/components/start_over_button_component.rb
@@ -1,0 +1,11 @@
+# frozen_string_literal: true
+
+# Overrides the default Blacklight StartOverButtonComponent in order to
+# always set the start_over_path the root path of the site/exhibit being searched
+class StartOverButtonComponent < Blacklight::StartOverButtonComponent
+  private
+
+  def start_over_path(_query_params = params)
+    params[:exhibit_id] ? helpers.exhibit_root_path(params[:exhibit_id]) : root_path
+  end
+end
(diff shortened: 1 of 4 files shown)


app/components/blacklight/start_over_button_component.rb
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+
+module Blacklight
```
jcoyne, line 3: Please don't use the namespace already in use in Blacklight (`Blacklight::StartOverButtonComponent`) . This is confusing. Instead, define `StartOverButtonComponent` at the root level (and extend `Blacklight::StartOverButtonComponent`)

jcoyne (changes requested):

app/components/blacklight/start_over_button_component.rb
dnoneill: If you don't use the blacklight namespace you can't override the StartOverButton in Blacklight.

jcoyne: I don't think you need to override it.  Make your own class.  call the new class in `app/views/catalog/_start_over.html.erb` This makes it very explicit about what you are doing.

dnoneill: Then we have to also override the constraints component as well. https://github.com/projectblacklight/blacklight/blob/d566efebb088ef87bffeb5a6b5a12c5017adda5f/app/components/blacklight/constraints_component.rb#L28

jcoyne: @dnoneill you can pass that class as an argument to ContstraintsComponent.  No overriding is required.

dnoneill: Thanks for the tip. Updated.

jcoyne (approved):

PR #2562, opened 2024-09-23
move search_form partials to components

### LINKED ISSUE #2556: search_fields is deprecated
```
DEPRECATION WARNING: search_fields is deprecated and will be removed from a future release (removed without replacement). (called from block in _app_views_shared__site_search_form_html_erb___1116619559615804202_119300 at /opt/app/exhibits/exhibits/releases/20240916110307/app/views/shared/_site_search_form.html.erb:3)
```

### LINKED ISSUE #2576: Blacklight::SearchBarPresenter is deprecated

dnoneill:
closes #2556 
closes #2576


app/components/site_search_form_component.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+class SiteSearchFormComponent < ViewComponent::Base
+  def presenter
+    Blacklight::SearchBarPresenter.new(controller, SearchAcrossController.blacklight_config)
+  end
+
+  def search_fields
```
dnoneill, line 8: Got this functionality from https://github.com/projectblacklight/blacklight/blob/d566efebb088ef87bffeb5a6b5a12c5017adda5f/app/components/blacklight/search_bar_component.rb#L53-L57

jcoyne: Should this class extend from `Blacklight::SearchBarComponent` and inherit this code?

dnoneill: I don't think so. Would be a lot of code we aren't using.

app/views/shared/_site_navbar.html.erb
```diff
@@ -4,15 +4,15 @@
...
+            <%= render SiteSearchFormComponent.new %>
           <% end  %>
         </li>
       </ul>
     <% else %>
       <ul class="navbar-nav mr-auto w-100 order-1 order-md-0 mt-3 mt-md-0">
         <li class="navbar-item">
           <% if Settings.feature_flags.search_across || in_search_across? %>
-            <%= render 'shared/old_site_search_form', presenter: Blacklight::SearchBarPresenter.new(controller, SearchAcrossController.blacklight_config) %>
+            <%= render OldSiteSearchFormComponent.new %>
```
jcoyne, line 15: Can we delete this whole clause? The feature flag is true in prod.  I think we can remove it now and get rid of all of this code.https://github.com/sul-dlss/shared_configs/blob/exhibits-prod/config/settings/production.yml#L30C3-L30C17

dnoneill: Double checked. It looks like we can. PR has removed this.

jcoyne (changes requested):
Can you also remove https://github.com/sul-dlss/exhibits/blob/main/config/settings.yml#L29 in this PR

dnoneill:
Good catch. Thanks.

spec/features/search_across_spec.rb
```diff
@@ -24,12 +24,18 @@
     tag1
   end
 
+  def js_search
+    within(first('.search-query-form')) do
+      find('#site-search-type').click
```
jcoyne, line 29: What is the purpose of these steps?  Can the be navigated in an accessible way (e.g. Found with text)?

jcoyne (changes requested):

spec/features/search_across_spec.rb
dnoneill: Removing the old site partial means by default the view looks like this on load:  Screenshot 2024-09-25 at 3 47 20 PM. There is no search button, so we need to click on the dropdown and choose Screenshot 2024-09-25 at 3 47 29 PM which does have a search button.

jcoyne: Can you do `click_on "Find exhibits by title"` and then `click_on "items in all exhibits"`?  That way we're also testing that this is accessible.

dnoneill: done.

app/components/site_search_form_component.rb
dnoneill: @jcoyne I didn't realize that the Blacklight::SearchBarPresenter was being deprecated. I have update the component to extend Blacklight::SearchBarComponent. Thanks for pointing me in the right direction.

jcoyne (approved):

dnoneill:
@jcoyne Do you mind doing another review? I had to move the _site_navbar content to a component to get the tests to work.

spec/components/site_navbar_component_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+      # rubocop:enable RSpec/AnyInstance
+    end
+
+    it { expect(rendered).to have_link('Request an exhibit') }
+    it { expect(rendered).not_to have_link('Create a new exhibit') }
+    it { expect(rendered).to have_content 'search form' }
+  end
+
+  context 'with an admin user' do
+    let(:user) { create(:admin) }
```
corylown, line 41: I think you want `site_admin` instead of `admin`

spec/components/site_navbar_component_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+
+    before do
+      # rubocop:disable RSpec/AnyInstance
+      allow_any_instance_of(ApplicationController).to receive(:can?).with([:create, :manage],
+                                                                          Spotlight::Exhibit).and_return(true)
+      allow_any_instance_of(ApplicationController).to receive(:can?).with(:create, Spotlight::Exhibit).and_return(true)
+      allow_any_instance_of(ApplicationController).to receive(:can?).with(:manage, Spotlight::Exhibit).and_return(true)
+      allow_any_instance_of(ApplicationController).to receive(:can?).with(:manage,
+                                                                          Spotlight::Site.instance).and_return(true)
+      # rubocop:enable RSpec/AnyInstance
```
corylown, line 51: I think this (as well as the setup for the previous test) could be simplified to something like:
```
allow(controller).to receive(:current_user).and_return(user)
```

corylown (changes requested):
@dnoneill this looks good to me. See comments about the test setup for `SiteNavbarComponent`.

spec/components/site_navbar_component_spec.rb
```diff
@@ -0,0 +1,32 @@
...
+
+RSpec.describe SiteNavbarComponent, type: :component do
+  subject(:rendered) do
+    Capybara::Node::Simple.new(render_inline(described_class.new).to_s)
+  end
+
+  before do
+    # rubocop:disable RSpec/AnyInstance
+    allow_any_instance_of(ApplicationController).to receive(:current_user).and_return(user)
+    # rubocop:enable RSpec/AnyInstance
```
corylown, line 13: Does this work so we don't have to disable `RSpec/AnyInstance`?
```suggestion
    allow(controller).to receive(:current_user).and_return(user)
```

dnoneill: Nope. Causes an error in CI

dnoneill: https://github.com/sul-dlss/exhibits/actions/runs/11146554130/job/30978782641

corylown: Oh, I think with `view_component` v3 `controller` needs to be `vc_test_controller`

dnoneill: undefined local variable or method `vc_test_controller' for #<RSpec::ExampleGroups::SiteNavbarComponent::WithANonAdminUser:0x00000001261b9910>

corylown: I think if you rebase this with what's currently in main you'll pick up view_components v3 and `vc_test_controller` will work. 🤞

corylown (approved):

Issue #2054, opened 2021-03-25
Parker on the Web: Everything search not picking up some bibliography data


blalbrit:
Per user:

The search box allows me to search keywords in bibliographic item titles but I can't search by the item's author - should that be possible and if not can we make it possible?

This refers to data pulled into Parker from Zotero - sample record here: https://parker.stanford.edu/parker/catalog/94257JYP

Search on keywords "scribal edition" returns a hit: https://parker.stanford.edu/parker/catalog?exhibit_id=parker&search_field=search&q=%22scribal+edition%22

Search on anything in the remaining fields is nil, for example: "Wood, Sarah" returns nothing - https://parker.stanford.edu/parker/catalog?exhibit_id=parker&search_field=search&q=%22Wood%2C+Sarah%22

Is it possible to index all content from the bibliography?

corylown:
@dnoneill the PR you link to https://github.com/sul-dlss/exhibits/pull/2551 addresses author search and https://github.com/sul-dlss/exhibits/pull/2552 will address searching of `*_ssim` fields. Do we need to also consider `pub_display` and `*_ssi` fields?

dnoneill:
That is a good point. Probably?

corylown:
pub_display still not searched. _ssi fields should be assessed for suitability for search.

hudajkhan:
A few comments:
- pub_display and pub_search are related, but may not be identical. pub_search is automatically added to all_search (which means a keyword search should return results), and pub_search is added to pub_display. (But pub_display is NOT added to pub_search).
- pub_display is populated based on bibtex_config.rb (extract_bibtex_field(:journal) and extract_bibtex_field(:publisher))
- pub_search is populated based on dor_config.rb (stanford_mods(:place))
- pub_search is also the field used for advanced search publisher text box (based on where it is mentioned in solrconfig.xml)

hudajkhan:
- ref_type_ssm (reference type) is populated based on bibtex_config (seems to be extracting zotero mapping)
- _ssm fields are stored multi-valued string fields but not indexed, which explains why they are not showing up in search.

hudajkhan:
The quickest mechanism to add both of these to search is to copy both ref_type_ssm and pub_display to all_search.  For pub_display, we can discuss if duplicates might show up in all_search (if pub_display ALSO has things that are already in pub_search) and if that is an issue. 

Tasks to complete:

- Code schema update
- Solr config schema update for staging
[...]

PR #2597, opened 2024-10-07
adding fields for searching for exhibits: pub_display and ref_type_ssm

hudajkhan:
Relates to #2054 .

What this PR does:
pub_search was already being added to pub_display, and we want pub_display to also be searchable.  This pull request populates pub_search with the journal and publisher info from bibtex as well as place info from dor configuration.  Schema.xml already includes a directive to copy pub_search to pub_display.  This enables searching across all the values that will be in pub_display.

The PR also copies ref_type_ssm to ref_type_ssim.  Schema.xml copies all *_ssim fields to all_search, so the values in this field will also be searchable. 

A few tests relied on pub_display values.  Since the code populates pub_search (and  the schema leads to copying of this field to pub_display), the PR updates those tests to use pub_search instead.


dnoneill:
I think you want to do this in the indexer no the schema.

hudajkhan:
@dnoneill To keep it in the code and not in the schema directly? ok I can try that instead.

hudajkhan:
Any *_search is automatically added to all_search so I can try that.

lib/traject/bibtex_config.rb
```diff
@@ -41,8 +41,8 @@
 to_fields %w(pub_year_isi pub_year_w_approx_isi), extract_bibtex(:year)
 to_field 'editor_ssim', extract_bibtex_field(:editor)
 to_field 'book_title_ssim', extract_bibtex_field(:booktitle)
-to_field 'pub_display', extract_bibtex_field(:journal)
-to_field 'pub_display', extract_bibtex_field(:publisher)
+to_field %w(pub_display pub_search), extract_bibtex_field(:journal)
+to_field %w(pub_display pub_search), extract_bibtex_field(:publisher)
```
corylown, line 45: Assuming we leave the solr schema as it was, these could both just change to:

```suggestion
to_field 'pub_search', extract_bibtex_field(:journal)
to_field 'pub_search', extract_bibtex_field(:publisher)
```

corylown (changes requested):

lib/traject/dor_config.rb
```diff
@@ -95,7 +95,7 @@
 to_field 'url_suppl', stanford_mods(:term_values, [:related_item, :location, :url])
 
 # publication fields
-to_field 'pub_search', stanford_mods(:place)
+to_field %w(pub_display pub_search), stanford_mods(:place)
```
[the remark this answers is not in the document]
hudajkhan: We want pub_display to be searchable, and as it stands, this would result in some circular logic.

Currently:

pub_search is copied to pub_display. In other words, pub_display is always pub_display's original values plus whatever is in pub_search.
pub_search is copied to all_search.

If we want pub_display to be searchable, we could copy it over pub_search. But pub_search is already copied over to pub_display.

If we want pub_display and pub_search to be equally searchable, it seems more consistent to make them equivalent from the beginning and remove the copy field in the schema.

lib/traject/bibtex_config.rb
hudajkhan: I think I understand your comments if we were to not make any changes to the schema at all, how might we streamline these values so everything gets added to pub_search (and then that gets copied to pub_display).  Maybe we could chat later.

corylown: In the current production Solr schema `pub_search` is getting copied to `pub_display`. There's a similar pattern in place for several other `_search` fields that get copied to `_display` fields. I'm proposing to leave this behavior as it is in the Solr schema since it's a pattern. For bibtext items add the journal and publisher info to `pub_search` and let the Solr schema copyfield directive copy `pub_search` to the `copy_display` field. This way we don't change the existing pattern, don't have to change the schema, and don't have to change anything about the dor handling. Happy to huddle on this.

dnoneill:
@hudajkhan My comments are exactly what Cory was suggesting.

corylown (approved):

PR #2607, opened 2024-10-14
Corylown update dependencies

corylown:


spec/controllers/spotlight/catalog_controller_spec.rb
```diff
@@ -22,6 +22,8 @@
         uploaded_resource.save_and_index
       end
 
+      sleep 10
```
dnoneill, line 25: @corylown I tried the sleep 10 in multiple spots of the code and it didn't work.

PR #2615, opened 2024-10-24
Update to Blacklight 8.

cbeer:


taylor-steve:
Here are two things I noticed. Not sure if they are a part of this work or not.

1. `/search?group=true&q=Senatorio&search_field=default&view=gallery` with "Group by exhibit" selected.
* 'Group by exhibit' isn't visually selected
* Missing exhibit count?

prod:
Screenshot 2024-11-08 at 2 01 09 PM

stage:
Screenshot 2024-11-08 at 2 01 00 PM

2. Same `/search?q=Senatorio&search_field=default` search but with "All results" selected. Should the IIIF icon be down below to the right like the other search results?

Screenshot 2024-11-08 at 2 32 43 PM

Great work. Happy to approve if these issues aren't relevant.

taylor-steve (approved):

Issue #2622, opened 2024-10-31
Don't display Parker specific fields on exhibits that are not Parker


corylown:

hudajkhan:
Based on the investigation issue, here are the five fields that only show up in Parker exhibits in Solr:
https://github.com/sul-dlss/exhibits/issues/2602 
"The fields that only showed up in the Parker exhibit were: book_title_ssim, editor_ssim, university_ssim, range_labels_ssim, and related_document_id_ssim."

corylown:
Closed by #2643

PR #2643, opened 2024-11-19
Display specific fields in the index and metadata configuration page only for Parker exhibit

hudajkhan:
Relates to #2622 .

What this PR does:
- Create a new feature flag for parker specific index fields
- Sets that flag to false in the main settings file. 
- In the catalog controller, only adds parker specific fields if that flag is set to true. 

How to test this:
- In local development, create an exhibit with the slug "parker".
- Set up Solr.  Add a single Solr document with the following fields indicating this is a parker exhibit:

```
("exhibit_parker_public_bsi": [
    true
  ],
  "spotlight_exhibit_slug_parker_bsi": [
    true
  ],
  "spotlight_exhibit_slugs_ssim": [
    "parker"
  ]
```
-  create a new settings.local.yml file (or similar). 
Set up:
feature_flags:
  parker:
    add_parker_index_fields: true

You should still see Parker specific fields in the metadata configuration page as well as the actual search results view, and you should not see Parker specific fields in any exhibit that is not Parker. 

I didn't see any specific feature flag tests around values i.e. add_parker_search_fields.  If anyone has thoughts on any tests to add, please feel free to suggest them.

Next step once this is merged: Updating feature flags for parker and parker-test here https://github.com/sul-dlss/shared_configs/blob/exhibits-prod/config/settings/production.yml so that the right values are picked up

### DIFF
--- a/app/controllers/catalog_controller.rb
+++ b/app/controllers/catalog_controller.rb
@@ -267,15 +267,23 @@ class CatalogController < ApplicationController
     config.add_index_field 'format_main_ssim', label: 'Resource type', if: lambda { |context, *_args|
       context.feature_flags.add_resource_type_index_field?
     }
+
+    # Parker specific fields (via Solr)
+    # editor_ssim, university_ssim, range_labels_ssim, and related_document_id_ssim.
+    config.add_index_field 'book_title_ssim', if: lambda { |context, *_args|
+      context.feature_flags.add_parker_index_fields?
+    }
+    config.add_index_field 'university_ssim', label: 'University'
+    config.add_index_field 'edition_ssm', label: 'Edition'
+    config.add_index_field 'range_labels_tesim', label: 'Section'
+    config.add_index_field 'related_document_id_ssim', label: 'Manuscript', helper_method: :manuscript_link
+
     # Fields added by Zotero API BibTeX import
     config.add_index_field 'volume_ssm', label: 'Volume'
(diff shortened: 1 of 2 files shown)


corylown (commented):
@hudajkhan this will need configurations to enable these fields for parker set in shared_config before this can be deployed to production, right? I think these flags should be set before we merge this work.

corylown (approved):

app/controllers/catalog_controller.rb
```diff
@@ -267,15 +267,32 @@ class CatalogController < ApplicationController
     config.add_index_field 'format_main_ssim', label: 'Resource type', if: lambda { |context, *_args|
       context.feature_flags.add_resource_type_index_field?
     }
+
+    # Parker specific fields we do not wish to show in other exhibits
+    # editor_ssim, university_ssim, range_labels_ssim, and related_document_id_ssim.
+    config.add_index_field 'book_title_ssim', if: lambda { |context, *_args|
```
corylown, line 273: @hudajkhan does this also need the label configured? It was set to "Book Title" below.

corylown (changes requested):

app/controllers/catalog_controller.rb
hudajkhan: Yes, I will add that

corylown (approved):

Issue #2445, opened 2024-07-03
Migrate from libsass to dart-sass


corylown:
Work in progress here: https://github.com/sul-dlss/exhibits/pull/2284

PR #2664, opened 2024-12-02
Adjust metadata display on browse list view

corylown:
Ideally we'd add the `col-xl-2` and `col-xl-10` classes in the template, but the only way to do that is to create a local version of `Blacklight::MetadataFieldComponent` that has to be configured for each field separately and then would need to be differently configured for the browse and catalog views. This css approach seemed less bad.

Before:
Screenshot 2024-12-02 at 4 50 51 PM

After:
Screenshot 2024-12-02 at 4 51 36 PM

### DIFF
--- a/app/assets/stylesheets/base.scss
+++ b/app/assets/stylesheets/base.scss
@@ -21,3 +21,4 @@
 @import "modules/caret";
 @import "modules/search_across";
 @import "modules/exhibit_search_typeahead";
+@import "modules/browse";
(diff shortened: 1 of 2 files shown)


app/assets/stylesheets/base.scss
```diff
@@ -21,3 +21,4 @@
 @import "modules/caret";
 @import "modules/search_across";
 @import "modules/exhibit_search_typeahead";
+@import "modules/browse";
```
jcoyne, line 24: `@import` is deprecated in SASS.  Do we want to continue to use this? https://sass-lang.com/blog/import-is-deprecated/

corylown: I think we need to get to `dart-sass` before we can stop using `@import`. We have an issue for it https://github.com/sul-dlss/exhibits/issues/2445

taylor-steve (approved):

PR #2675, opened 2024-12-05
Use the new header layout

### LINKED ISSUE #2594: Exhibits header redesign
_Implement redesigned exhibits header for improved accessibility and design system alignment._

The Stanford Office of Digital Accessibility (SODA) confirmed the following accessibility issues with the current exhibits header design. Text appearing overlaid on background images must comply with WCAG color contrast guidelines. The masthead background images are specified by exhibit creators, and currently do not always satisfy the color contrast ratio with the exhibit title text.

Design and Engineering are working to align the header design across applications via the design system [component library](https://github.com/sul-dlss/component-library). The exhibits header poses unique challenges to this effort because it contains elements that do not appear in other applications. These elements include: a background masthead image, an exhibit subtitle ([sometimes](https://exhibits.stanford.edu/vabamu) [lengthy](https://exhibits.stanford.edu/publicsyncopations)), header dropdown options ([sometimes lengthy](https://exhibits.stanford.edu/martin-wong)).

Because these elements are specified by exhibit creators, the ideal exhibit header design should satisfy accessibility requirements regardless of the creators’ choices.

The exhibit header redesign options are available in the “Exhibits Redesign 2024” Figma File: https://www.figma.com/design/0LTbMiP5xdL6T94nmlKDci/Exhibits-Redesign-2024?node-id=0-1&t=8KDvdYTqfPLMdHwC-1
Image

The base header component can be found in “DLSS Components” Figma File: https://www.figma.com/design/K1syTsalaB1nlRftgOi0Pw/DLSS-Components?node-id=742-922&t=hezVZDhbDPM3h5IU-4
**NB**: The `Application Subtitle Notes` for responsive font size and display.
Image

jcoyne:
Screenshot 2024-12-05 at 4 48 59 PM

Fixes #2594


alundgard:
Looking good so far! Here are a few differences I noticed right away.
Comparison

It will help to see the new header on stage so we can look at exhibits other than the homepage. Such as these pages.
https://www.figma.com/design/0LTbMiP5xdL6T94nmlKDci/Exhibits-Redesign-2024?node-id=0-1&m=dev&focus-id=236-6080

alundgard:
Now that the search bar is on the left side of the page, we'd like the Search tips button to move to the right side of the bar.
Comparison-page

alundgard:
The header background color should be Stanford Fog Light `#F4F4F4`. It is currently something lighter than that.

corylown:
This looks great! A couple minor things I noticed -- some of which may be questions for @alundgard 

- The order of the links in `_user_util_links` doesn't match the design and the design doesn't account for the case where a language picker is present.
  - See https://exhibits-stage.stanford.edu/vt-test as an example
  - Design specifies: Spotlight at Stanford => Feedback => Login
  - Current order is: Language picker => Spotlight at Stanford => Login => Feedback
  - I think Spotlight at Stanford should always be in the leftmost position and login always on the far right, but wonder what @alundgard  thinks?
Screenshot 2024-12-06 at 2 29 21 PM



- Typeahead dropdown width is too narrow
  - This is only on the home page: https://exhibits-stage.stanford.edu/
Screenshot 2024-12-06 at 2 29 49 PM



- @alundgard should we be doing something different with these search-related buttons now that we're moving away from the cardinal red accents?
Screenshot 2024-12-06 at 2 43 03 PM

app/components/exhibit_navbar_component.html.erb
```diff
@@ -0,0 +1,13 @@
+<div id="site-navbar" class="site-navbar mt-auto py-2">
+    <% if helpers.should_render_spotlight_search_bar? %>
+      <div class="container d-flex">
+        <%= render search_component.new(
+          url: helpers.search_action_url,
+          advanced_search_url: helpers.search_action_url(action: 'advanced_search'),
+          params: helpers.search_state.params_for_search.except(:qt),
+          autocomplete_path: suggest_index_catalog_path
+        ) %>
+        <%#<%= render SearchTipsLinkComponent.new %>
```
corylown, line 10: We want search tips to display. Looks like it was commented out maybe unintentionally.

jcoyne: Yes. That was unintentional. It's been readded.

jcoyne:
@corylown typeahead and moving "Spotlight at Stanford" to the first link have been done.

corylown:
@alundgard a separate ticket for adopting the component library for the search-related button sounds right to me.

jcoyne:
@alundgard  I don't know a way for us to separate this out from the component library.  I think we have to pull off the bandaid and completely adopt the component library here or we have to create a bunch of intermediate code that we have to test just for this change, which will later have to be removed when fully committing to the component library.

alundgard:
It looks like the responsive font sizing is applied to the exhibit title, but not to the subtitle? I think we are mostly concerned to make the subtitle responsive since it is in-line with the menu options.

See https://exhibits-stage.stanford.edu/header-redesign-test-exhibit/

### Current
current-subtitle

### Goal
goal-subtitle

### Considerations

The subtitle should also have a _minimum_ font-size (~16px) so that it doesn't continue shrinking beyond legibility. (Currently, the subtitle isn't displayed for narrow viewports, but I can make a separate issue to address this aspect, if we have time.)
DLSS App Header v 2 - Subtitle

alundgard:
Here are a few more suggestions. You may have better ideas about how to address these changes.

- Change header base font-size from `0.9375rem` to `1rem` (as in the component library header)?

- Reduce margin-bottom on `site-title` class to bring the subtitle a bit closer (eg from `0.5rem` to `0.1rem`).

- Close small gap between dropdown menu. Adjust y-offset on `dropdown-menu` class from 39px to ~35px.

- Reduce height or padding on `search-across-form` class to make the search box a little narrower (eg reduce padding from `0.5rem` to `0.4rem`).
Comparison - Before After

- Double check the font-weight on the title. It's currently set to `500` but it appears no different from `400`? I think we'd like it be a bit heavier (see design below), but setting it `600` in the browser appears to be too much.
Comparison-page

alundgard:
When an exhibit doesn't have a subtitle, we'd still like the menu options to appear on the right side.
Example: https://exhibits-stage.stanford.edu/david-rumsey-map-collection

### Current
no-subtitle-current

### Desired
no-subtitle-desired

alundgard:
This may have been addressed since I last checked this branch on stage. It looks like including the `Search tips` button throws off the mobile search bar alignment. For narrow viewports, we'd like the `Search tips` button to wrap so that it appears below the search bar.

### Current
search-tips-alignment-current

### Design
Width=BS 5 XS (330px) - Menu Closed, Theme=Light

alundgard:
Is there a way to omit those `.btn-secondary` changes from this branch before merging? I created a separate ticket for the Search-control style update.

jcoyne:
@alundgard I've got a temporary fix ready that can go right after this.

corylown (approved):

PR #2680, opened 2024-12-06
Remove blacklight hierarchy

taylor-steve:
Moves away from blacklight-hierarchy as suggested in https://github.com/sul-dlss/exhibits/pull/2549.

Makes it easier to get Exhibits onto Spotlight v5.
Screenshot 2024-12-06 at 11 34 35 AM


app/components/blacklight/name_roles_facet_hierarchy_component.rb
```diff
@@ -0,0 +1,41 @@
+# frozen_string_literal: true
+
+module Blacklight
```
corylown, line 3: I'm not sure this should be name-spaced under `module Blacklight`. It makes it seem like an override when it's a local class that inherits behavior from `Blacklight::FacetFieldListComponent`.

corylown (changes requested):

app/components/blacklight/name_roles_facet_hierarchy_component.rb
taylor-steve: True, thanks. I was over-exuberant borrowing from https://github.com/sul-dlss/earthworks/blob/main/app/components/blacklight/date_facet_hierarchy_component.rb

corylown (approved):

PR #2682, opened 2024-12-06
For item that is a reference, fixes the list of documents which include that reference

### LINKED ISSUE #2673: Parker: Bib item homepage display issue
Reported by Ben Albritton today:

On the Bibliographic Item homepage we have a new issue as well. For example:
https://parker.stanford.edu/parker/catalog/3TZUP2CU 
In the section that is headed "This reference is found in these document bibliographies", we should only see the manuscripts that are mentioned in the bibliographical item (in this case, MSS 002I, 002II, 002III, 003, 053, and 020). Instead, there seems to be a large number of page-level information and perhaps the entire bibliography for the manuscripts as well. Probably just not filtering out the manuscript items from all of the other data

hudajkhan:
Closes #2673 .

Background:

- When a page like https://parker.stanford.edu/parker/catalog/3TZUP2CU is brought up, the "this reference is found in these document bibliographies" section retrieves a list of ids from the "related_document_id_ssim" field (e.g. for 3TZUP2CU), and then generates a list of links to those specific items.  
- The original query was querying the catalog's JSON API directly by sending in an ORed query string with the ids. This query was not limiting to the id field and was effectively a keyword search that could return multiple documents which were not specifically the ones indicated by those ids. 

What this pull request does:
- Sets up another query mechanism, relying on the search service fetch method when given an array of ids, which retrieves the documents with those specific ids.  
- On the Solr side, this appears to be generating a query of this format: "q"=>"{!lucene}id:(wz026zp2442 OR W4DJWZ9C)".  If we query Solr directly through its interface, this appears equivalent: q=id:(wz026zp2442 OR W4DJWZ9C).

### DIFF
--- a/app/controllers/catalog_controller.rb
+++ b/app/controllers/catalog_controller.rb
@@ -487,7 +487,7 @@ class CatalogController < ApplicationController
   def documents_list
     search_service = Blacklight::SearchService.new(config: blacklight_config)
     ids = params[:ids].present? ? params[:ids].split : []
-    @documents = ids.empty? ? [] : search_service.fetch(ids)
+    @documents = ids.empty? ? [] : search_service.fetch(ids, { :rows => 1000 })
     render json: @documents
   end
 
--- a/app/controllers/catalog_controller.rb
+++ b/app/controllers/catalog_controller.rb
@@ -487,8 +487,6 @@ class CatalogController < ApplicationController
   def documents_list
     search_service = Blacklight::SearchService.new(config: blacklight_config)
     ids = params[:ids].present? ? params[:ids].split : []
-    puts "DOCUMENTS LIST IDS"
-    puts ids.inspect
     @documents = ids.empty? ? [] : search_service.fetch(ids)
     render json: @documents
   end
--- a/app/controllers/catalog_controller.rb
+++ b/app/controllers/catalog_controller.rb
@@ -484,6 +484,15 @@ class CatalogController < ApplicationController
     end
   end
 
+  def documents_list
+    search_service = Blacklight::SearchService.new(config: blacklight_config)
+    ids = params[:ids].present? ? params[:ids].split : []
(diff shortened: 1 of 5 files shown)


app/views/catalog/_cited_documents_default.html.erb
```diff
@@ -1,7 +1,7 @@
 <% if @document.reference? && @document.cites_other_documents? %>
   <div class='col-md-12 record-metadata-section'
        data-behavior='cited-documents-contents'
-       data-path="<%= search_exhibit_catalog_path(exhibit_id: @exhibit) %>"
+       data-path="<%= url_for([@spotlight, :documents_list]) %>"
```
corylown, line 4: What's the purpose of `@spotlight` here? I think `url_for(:documents_list)` ought to work.

app/controllers/catalog_controller.rb
```diff
@@ -484,6 +484,13 @@ def search_tips
     end
   end
 
+  def documents_list
+    search_service = Blacklight::SearchService.new(config: blacklight_config)
+    ids = params[:ids].present? ? params[:ids].split : []
+    @documents = ids.empty? ? [] : search_service.fetch(ids)
```
corylown, line 490: Should we set a limit on the number of docs to fetch from Solr at once? Maybe 1000?

corylown (changes requested):

app/controllers/catalog_controller.rb
hudajkhan: The original approach did use 1000 as a cap, so we can do that here as well

app/views/catalog/_cited_documents_default.html.erb
hudajkhan: Correct, I changed it

corylown (approved):

PR #2694, opened 2024-12-10
Simplify styles on full text matches

jcoyne:


app/assets/stylesheets/modules/full_text_highlight.scss
```diff
@@ -3,10 +3,8 @@ dt.prepared-search-container {
   color: $gray-dark;
   cursor: pointer;
   display: inline-block;
-  float: none;
```
jcoyne, line 6: You can't float flex items.

app/assets/stylesheets/modules/full_text_highlight.scss
```diff
@@ -3,10 +3,8 @@ dt.prepared-search-container {
...
   text-align: initial;
   width: auto;
-  @extend .col-12;
```
jcoyne, line 9: This already has .col-3 on the element and since we have `width: auto` this adds nothing.

app/assets/stylesheets/modules/full_text_highlight.scss
```diff
@@ -17,10 +15,9 @@ dt.prepared-search-container {
 
 dd.blacklight-full_text_tesimv {
   border-left: 1px solid $color-gray-80;
-  float: none;
   margin-left: 15px;
   margin-top: 6px;
-  @extend .col-12;
```
jcoyne, line 23: This node already has a .col class on it, so all we need to do is set the width.

corylown (approved):

PR #2714, opened 2024-12-11
Remove range limit stylesheet

jcoyne:
Because we aren't using the histogram, this is unnecessary


taylor-steve:
It does subtly alter the spacing.

With:
Screenshot 2024-12-11 at 8 04 28 AM

Without:
Screenshot 2024-12-11 at 8 04 52 AM

Just something I noticed yesterday during the v5 migration. Not suggesting we can't drop it.

jcoyne:
Yes, I noticed that, but I think the tighter spacing is preferable.

taylor-steve:
Me too.

taylor-steve (approved):

PR #2752, opened 2024-12-17
add image selection for item embed viewers

### LINKED ISSUE #2604: Enable image area selection for Item Embed - Mirador
The Item Embed widget allows creators to search and select an image object from a dropdown list of items currently in the exhibit. Upon selection of the image object, there appear additional options: `Change` to select a different image if the object has multiple images, and `Remove` to start over.

Add a third option to the above two: `Select image area` to specify the region of interest.

Clicking `Select image area` opens a modal for the “cropping” interface.
Image
Image

### Syncing view preview and Change drawer
- For Stanford Digital Repository items with multiple images, the interface currently shows a `Image 1 of X [Change]` drawer for selecting the image to embed. This drawer functionality should sync with the modal selection.
- For example, if a user selects the 2nd image in the drawer, then opens the `Select image area` modal, the 2nd image should appear in the viewer embedded in the modal.
- Likewise, if a user selects and Saves the 2nd image in the modal, `Image 2 of X` should appear as selected in the `Change` drawer after they close the modal.
Image
Image

### Syncing view preview and column width
- When creators embed an image, they have the option to include text that appears as a second column on the left/right side of the viewer. If they chose to include this text, the viewer width is halved.
- If possible, the width of the viewer embedded in the `Select image area` modal should adjust depending on whether the creator includes text (halve the viewer width if text is included). This way the creator has a viewer preview that accurately represents how their embedded item will appear in the exhibit.

dnoneill:
closes #2604.
Needs https://github.com/sul-dlss/sul-embed/pull/2279 to work.


config/initializers/oembed.rb
```diff
@@ -2,9 +2,11 @@
 
 OEmbed::Providers.register_all
 
-purl_provider = OEmbed::Provider.new('http://purl.stanford.edu/embed.{format}?&hide_title=true')
+#purl_provider = OEmbed::Provider.new('http://purl.stanford.edu/embed.{format}?&hide_title=true')
+purl_provider = OEmbed::Provider.new('https://sul-purl-stage.stanford.edu/embed.{format}?&hide_title=true')
```
dnoneill, line 6:
```suggestion
purl_provider = OEmbed::Provider.new('https://sul-purl-stage.stanford.edu/embed.{format}?&hide_title=true')
purl_provider = OEmbed::Provider.new('http://localhost:3001/embed.{format}?&hide_title=true')
```

for testing locally

config/settings.yml
```diff
@@ -30,7 +30,7 @@ feature_flags:
 traject:
   processing_thread_pool: 1
 iiif_embed:
-  url: https://embed.stanford.edu/iiif
+  url: https://embed-stage.stanford.edu/iiif
```
dnoneill, line 33:
```suggestion
  url: https://embed-stage.stanford.edu/iiif
  url: http://localhost:3001.stanford.edu/iiif
```
For testing locally

app/assets/javascripts/request_viewer_state.js
```diff
@@ -0,0 +1,93 @@
...
+        return urlObj.href;
+      },
+
+      requestState: function() {
+          const iframe = document.querySelector('.oembed-widget iframe, iframe.mirador-embed-wrapper');
+          iframe.contentWindow.postMessage(JSON.stringify({ type: 'requestState' }), '*'); // Change '*' to a specific origin for security?
+      }
+  };
+
+  global.RequestViewerState = RequestViewerState;
```
jcoyne, line 84: Why do we need to set this? Can we just import the `init` function?

app/assets/javascripts/request_viewer_state.js
```diff
@@ -0,0 +1,93 @@
+/* eslint-disable camelcase */
+/* global Blacklight */
+
+(function (global) {
+  var RequestViewerState;
```
jcoyne, line 5: Why do we need a var?  can you just put const on the next line?

app/assets/javascripts/select_image_area.js
```diff
@@ -0,0 +1,63 @@
+/* eslint-disable camelcase */
+/* global Blacklight */
+
+(function (global) {
+  var SelectImageArea;
```
jcoyne, line 5: why does this need to be a var?  can we just put const on the next line?

app/assets/javascripts/select_image_area.js
```diff
@@ -0,0 +1,63 @@
...
+      const image_url = this.panel[0].querySelector('img');
+      if (image_url.src.includes('!33')) {
+        image_url.insertAdjacentHTML('afterend', '<span id="page-2-placeholder">This section spans two pages, we can not display the thumbnail for page 2.</span>');
+      }
+
+      target.before(selectImageAreaHtml);
+    }
+  };
+
+  global.SelectImageArea = SelectImageArea;
```
jcoyne, line 34: Can we just import the init function?

app/assets/javascripts/spotlight.js
```diff
@@ -1,2 +1,4 @@
 //= require spotlight/application
 //= require sir_trevor_block_overrides
+//= require select_image_area
```
jcoyne, line 3: Are we able to do this without continuing to use Sprockets?  Sprockets is deprecated, so that just seems to dig us deeper into debt.

app/assets/javascripts/select_image_area.js
```diff
@@ -0,0 +1,63 @@
...
+}(this));
+
+Blacklight.onLoad(function () {
```
jcoyne, line 37: Is there a reason we need to depend on `Blacklight` here?

app/assets/javascripts/request_viewer_state.js
```diff
@@ -0,0 +1,93 @@
...
+          const iframe = document.querySelector('.oembed-widget iframe, iframe.mirador-embed-wrapper');
+          iframe.contentWindow.postMessage(JSON.stringify({ type: 'requestState' }), '*'); // Change '*' to a specific origin for security?
+      }
+  };
+
+  global.RequestViewerState = RequestViewerState;
+}(this));
+
+
+Blacklight.onLoad(function () {
```
jcoyne, line 88: Is there a reason we need to depend on `Blacklight` here? Won't this create a duplicate event listener on every pageview (assuming turbo or turbolinks is used).

app/assets/javascripts/select_image_area.js
```diff
@@ -0,0 +1,63 @@
...
+          if (node.nodeType === 1 && node.tagName === 'LI') {
+            const isEmbed = node.closest('[data-type]').dataset.type == 'solr_documents_embed';
+            if (isEmbed) SelectImageArea.init(node);
+          }
+        });
+      }
+    }
+  };
+
+  document.querySelectorAll('#page-content').forEach(function(element, i) {
```
jcoyne, line 58: Why do we need a `querySelectorAll` on an id?  Id should be unique, so `getElementById` should be used.

jcoyne (changes requested):

hudajkhan:
I tried this out with the sul-embed branch, and this feature works as intended.

app/assets/javascripts/request_viewer_state.js
dnoneill: This seems to be how we are doing it in all our JS. If you have specific code I can plug in to test to see if it works then I will do that, but otherwise I do not think it is worth the time it will take me to figure it out. Especially since this is inline with what we are doing in this application. 

https://github.com/sul-dlss/exhibits/blob/main/app/assets/javascripts/bibliography.js#L101
https://github.com/sul-dlss/exhibits/blob/main/app/assets/javascripts/nested_related_items.js#L86

app/assets/javascripts/request_viewer_state.js
dnoneill: Nope. cruft from before we figured out the blacklight-modal event

app/assets/javascripts/spotlight.js
hudajkhan: That sounds correct @dnoneill

app/assets/javascripts/select_image_area.js
dnoneill: see above

app/assets/javascripts/select_image_area.js
dnoneill: doesn't work without it.

app/assets/javascripts/request_viewer_state.js
```diff
@@ -0,0 +1,88 @@
+/* eslint-disable camelcase */
+/* global Blacklight */
```
jcoyne, line 2: Where is Blacklight used here?

app/assets/javascripts/select_image_area.js
```diff
@@ -0,0 +1,62 @@
+/* eslint-disable camelcase */
+/* global Blacklight */
```
jcoyne, line 2: Where is Blacklight used here?

app/assets/javascripts/request_viewer_state.js
jcoyne: All of this JS was written 10 years ago, pre ES2015 standard.  Just because it's been that way in the past, doesn't mean we should repeat those patterns.

app/assets/javascripts/select_image_area.js
jcoyne: Blacklight.onLoad is just a thin wrapper around `document.addEventListener("DOMContentLoaded", ...`

app/assets/javascripts/spotlight.js
jcoyne: I'm pretty sure these can be loaded independently (via importmap), since they have no dependency on anything within Spotlight.

app/views/catalog/_embedded_mirador3.html.erb
```diff
@@ -1,7 +1,7 @@
 <% if document.manifest_url.present? %>
   <% manifest_url = document.manifest_url.starts_with?('/') ? root_url + document.manifest_url.slice(1..-1) : document.manifest_url %>
   <%= content_tag :iframe, '',
-        src: "#{Settings.iiif_embed.url}?#{{ url: manifest_url }.to_query}",
+        src: "#{Settings.iiif_embed.url}?#{{ url: manifest_url, canvas_id: choose_canvas_id(block) || params[:canvas_id], iiif_initial_viewer_config: choose_initial_viewer_config(block) || params[:iiif_initial_viewer_config]}.to_query}",
```
jcoyne, line 4: This is a lot of code for an erb file.  Can this be turned into a ViewComponent, so that some of this can be plain ruby methods?

app/controllers/catalog_controller.rb
```diff
@@ -504,6 +504,25 @@ def search_tips
     end
   end
 
+  def select_image_area
```
jcoyne, line 507: This controller is already doing too many things.  Please move this to a new controller that is about "image areas".

config/initializers/oembed.rb
```diff
@@ -2,9 +2,11 @@
 
 OEmbed::Providers.register_all
 
-purl_provider = OEmbed::Provider.new('http://purl.stanford.edu/embed.{format}?&hide_title=true')
+#purl_provider = OEmbed::Provider.new('http://purl.stanford.edu/embed.{format}?&hide_title=true')
```
jcoyne, line 5: Should this be reverted?

config/settings.yml
jcoyne: The place for local values is  `config/settings/developement.yml`

app/assets/javascripts/request_viewer_state.js
```diff
@@ -0,0 +1,88 @@
...
+                try {
+                    parsedData = typeof event.data === 'string' ? JSON.parse(event.data) : event.data;
+                } catch (error) {
+                    console.error('Failed to parse event data:', error);
+                    return; // Exit if parsing fails
+                }
+
+                if (parsedData.type === "stateResponse" && parsedData.source === "sul-embed-m3") {
+
+                    console.log('Exhibits: received state:', parsedData.data);
```
jcoyne, line 35: Please extract this block to a separate method. The outer method is very long and hard to fully understand.

app/assets/javascripts/request_viewer_state.js
```diff
@@ -0,0 +1,88 @@
...
+                    let { viewers, windows, iiif_images, canvas_index } = parsedData.data;
+                    if (document.querySelector('#page-2-placeholder')) document.querySelector('#page-2-placeholder').remove();
+                    const viewer = Object.keys(viewers)[0];
+                    const canvas_id = windows[viewer]['canvasId'];
+                    const iiif_initial_viewer_config = JSON.stringify(viewers[viewer]);
+                    const canvasField = document.querySelector(`#${this.formId} > input[name="item[${this.itemId}][iiif_canvas_id]"]`)
```
jcoyne, line 41: Can you do 
```
const formNode = document.getElementById(this.formId);
formNode.querySelector("input[name="item[${this.itemId}][iiif_canvas_id]")
```

app/assets/javascripts/request_viewer_state.js
```diff
@@ -0,0 +1,88 @@
...
+                }
+
+                if (parsedData.type === "stateResponse" && parsedData.source === "sul-embed-m3") {
+
+                    console.log('Exhibits: received state:', parsedData.data);
+                    let { viewers, windows, iiif_images, canvas_index } = parsedData.data;
+                    if (document.querySelector('#page-2-placeholder')) document.querySelector('#page-2-placeholder').remove();
+                    const viewer = Object.keys(viewers)[0];
+                    const canvas_id = windows[viewer]['canvasId'];
+                    const iiif_initial_viewer_config = JSON.stringify(viewers[viewer]);
```
jcoyne, line 40: Use camel case variable names for javascript.

jcoyne (changes requested):

config/initializers/oembed.rb
dnoneill: Yeah the stage stuff is just for deploying to stage for users to test.

app/views/catalog/_embedded_mirador3.html.erb
dnoneill: I think that is outside the scope of this very large PR. Can move this into a helper method but I think moving this into a ViewComponent should be it's own PR. I think if we are going to move this into a ViewComponent we should also move https://github.com/sul-dlss/exhibits/blob/dff44d8436cec0e9d950f4bf501b1228a03b9c4a/app/views/catalog/_oembed_default.html.erb into a ViewComponent. So I think that should be it's own PR.

config/initializers/oembed.rb
jcoyne: Can you put this in a config/settings.yml and use shared config for this?

app/assets/javascripts/request_viewer_state.js
dnoneill: If you have specific code I can plug in to test to see if it works then I will do that, but otherwise I do not think it is worth the time it will take me to figure it out. We only have til the end of this week to get this into exhibits and it is a priority.

app/assets/javascripts/select_image_area.js
dnoneill: Are you suggesting using document.addEventListener("DOMContentLoaded" instead?

app/assets/javascripts/spotlight.js
taylor-steve: @jcoyne is correct, we could definitely handle this with an importmap now, it works fine with sprockets and isn't dependent on Spotlight. 

We could also incorporate this into the upcoming jsbundling work, I think that'd be a fine approach as well.

It looks like we might want to update the JS slightly for either approach, I will add to those specific discussions.

app/controllers/catalog_controller.rb
dnoneill: This is taking more than an hour to figure out. I am getting can can errors, etc.

app/assets/javascripts/select_image_area.js
dnoneill: It doesn't work. I tried multiple iterations of DOMContentLoaded and it didn't work. I am guessing it is due to Spotlight waiting for Blacklight to load and this code is dependent on Spotlight JS running.
https://github.com/projectblacklight/spotlight/blob/5384fca74a40faf5e214900afab44733ed67fb6e/lib/generators/spotlight/assets/importmap_generator.rb#L62-L64

PR #2754, opened 2024-12-17
Enable skip links to show above the fold in their entirety

### LINKED ISSUE #2473: Selecting view type or filter should not shift focus to the top of the page
### Current behavior

Selecting a view type or filter using the keyboard or screen reader shifts the focus back to the top of the page.

### Steps to reproduce

Example page: [Spotlight Accessibility for SODA - San Francisco Buildings](https://exhibits.stanford.edu/a11y-for-soda/browse/san-francisco-buildings).

Navigate to the view type buttons with the keyboard or screen reader and make a selection. Also, navigate to the sidebar filters and make a selection.

### Possible solution

WCAG Level A: [Understanding Success Criterion 3.2.1: On Focus](https://www.w3.org/WAI/WCAG21/Understanding/on-focus.html).

SODA recommends moving the focus to the beginning of the updated content (in this case, the exhibit items). Doing so would mean jumping over the Search input field, which sits between the view options and exhibit items. Or, when selecting a filter from the sidebar, focus would jump across sections of the page.

Another possibility may be to leave focus in its current position (at the view type button or sidebar filter), but with an indication that it has been selected (see #2472).

hudajkhan:
Closes #2473 .

To be more in line with the horizontal layout of all the links, and to reduce overrides, the following PR does the following:
- overrides the #skip-link position:absolute CSS property inherited from Spotlight, which allows the section to display above the header, similar to how EarthWorks currently operates. 
- removes the custom skip link component and overridden file, since they are no longer required.  Removes the catalog controller setting which uses the custom skip link component.

Screenshot 2024-12-17 at 2 27 22 PM


jcoyne (changes requested):
This seems to me like a step backward.  We want to use the style defined in the component-library, not the out of date style from Blacklight 8.7, which will be removed in Blacklight 9.

jcoyne:
It's possible that we do want to make changes in the component library as the focus ring doesn't seem to work great when the header background color is set:
Screenshot 2024-12-17 at 3 46 37 PM

app/assets/stylesheets/modules/spotlight_overrides.scss
```diff
@@ -220,3 +220,9 @@ h5,
     z-index: 1;
   }
 }
+
+// This allows for the skip links to appear above the header
+// instead of overlapping with the header. 
+#skip-link {
```
jcoyne, line 226: Should this change be in the component library?

app/components/skip_link_component.html.erb
```diff
@@ -1,5 +1,5 @@
-<nav id="skip" class="visually-hidden-focusable" aria-label="<%= t('blacklight.skip_links.label') %>">
+<nav id="skip-link" class="visually-hidden-focusable" aria-label="<%= t('blacklight.skip_links.label') %>">
```
jcoyne, line 1: This was intentionally made to not be `#skip-link` so that we didn't have to fight against the CSS in blacklight.

hudajkhan: The CSS for #skip-link is coming in from spotlight, with really just one change required for the position to make it appear correctly

app/assets/stylesheets/modules/spotlight_overrides.scss
hudajkhan: We can put it in there to ensure this overrides any absolute positioning, since that does seem to interfere.

app/components/skip_link_component.html.erb
jcoyne: Right, but we want all of the dlss sites to have the same skip link look, even if they aren't using Spotlight.

hudajkhan:
@jcoyne @alundgard it would be good to get your feedback here.  

For reference, here's what the skip links look like in the component library site:
Screenshot 2024-12-20 at 10 31 19 AM

If I use the skip-link implementation while applying the skip-link style from Spotlight (and overriding just the position:absolute) part, this is what it looks like

Screenshot 2024-12-20 at 10 35 28 AM

If I change the id so that the Spotlight styling is not applied, we get this
Screenshot 2024-12-20 at 10 36 45 AM

alundgard:
@hudajkhan I think any of the above screenshots are acceptable. The focus ring overlaying the Stanford banner is not ideal, but it shouldn't affect usability. It's more important to ensure that the skip links do what they're supposed to do.

jcoyne (approved):

hudajkhan:
For reference, the links now look like this:
Screenshot 2024-12-20 at 2 50 04 PM

PR #2758, opened 2024-12-18
Add visually-hidden-focusable to nav element

### LINKED ISSUE #2756: Skip links visible on search results pages
https://exhibits.stanford.edu/data/catalog?exhibit_id=data&search_field=search&q=

Image

corylown:
Fixes #2756 

I've lost the plot on what we're doing with skip links...but adding `visually-hidden-focusable` to the `nav` element fixes a production issue where the `Skip to first result` link displays on search results pages. Open to other fixes. This appears to be what Blacklight 8 does: https://github.com/projectblacklight/blacklight/blob/ab210bd3d5cf352d10a2a6052970d34306a6d17f/app/components/blacklight/skip_link_component.html.erb#L1

### DIFF
--- a/app/components/skip_link_component.html.erb
+++ b/app/components/skip_link_component.html.erb
@@ -1,4 +1,4 @@
-<nav id="skip" aria-label="<%= t('blacklight.skip_links.label') %>">
+<nav id="skip" class="visually-hidden-focusable" aria-label="<%= t('blacklight.skip_links.label') %>">
     <%= link_to_main %>
     <%= link_to_search %>
     <%= content %>


app/components/skip_link_component.html.erb
```diff
@@ -1,4 +1,4 @@
-<nav id="skip" aria-label="<%= t('blacklight.skip_links.label') %>">
+<nav id="skip" class="visually-hidden-focusable" aria-label="<%= t('blacklight.skip_links.label') %>">
```
jcoyne, line 1: Nav should not have this class.  It belongs on the anchor tag. See https://github.com/sul-dlss/component-library/blob/main/header/index.html#L34

corylown: Maybe this is a blacklight bug then? I think the skip to first result link is coming from blacklight: https://github.com/projectblacklight/blacklight/blob/6a5f890ccbd379085c4bd74363619650f3872764/app/views/catalog/_search_results.html.erb#L11

jcoyne: Yes. Exactly.  we need to put the classes on that link.

corylown: We can work on fixing this in Blacklight, but it seems not ideal to have this in production in exhibits until that gets resolved upstream. The other option I see is override the whole search_results partial in exhibits.

As part of the Blacklight fix should we backport https://github.com/projectblacklight/blacklight/pull/3461 to Blacklight 8?

PR #2763, opened 2024-12-18
fix search bar for small screens

### LINKED ISSUE #2721: Header: Search bar alignment adjustments
### Fix vertical alignment of search dropdown
The `Find exhibits by title` dropdown appears to be vertically misaligned on https://exhibits-stage.stanford.edu/.
Image

The same dropdown looks okay on the individual exhibit pages, such as https://exhibits-stage.stanford.edu/david-rumsey-map-collection.
Image

### Adjust search icon spacing
The horizontal alignment of the search icon appears off. It looks like there is too much padding on the right side, and not enough on the left side. Can we make the icon more centered in the available space, so that it isn't as close to the edge of the search bar.
Image

dnoneill:
closes #2721

### DIFF
--- a/app/components/site_search_form_component.html.erb
+++ b/app/components/site_search_form_component.html.erb
@@ -1,6 +1,6 @@
 <%= form_tag @url, method: :get, class: 'search-across-form search-query-form row gx-0', role: 'search', 'aria-label' => t('blacklight.search.form.submit') do %>
-  <div class="col-auto search-q"></div>
-  <div class="col-auto">
+  <div class="col-auto search-q d-none d-sm-block"></div>
+  <div class="col-auto align-self-center">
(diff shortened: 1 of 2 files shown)


app/components/site_search_form_component.html.erb
```diff
@@ -1,6 +1,6 @@
-<%= form_tag @url, method: :get, class: 'search-across-form search-query-form row gx-0', role: 'search', 'aria-label' => t('blacklight.search.form.submit') do %>
-  <div class="col-auto search-q"></div>
-  <div class="col-auto">
+<%= form_tag @url, method: :get, class: 'search-across-form search-query-form row gx-0 flex-nowrap', role: 'search', 'aria-label' => t('blacklight.search.form.submit') do %>
```
jcoyne, line 1: Is `flex-nowrap` required here?  Isn't that the default?

dnoneill: It isn't the default, but this is cruft from another solution. Good catch.

jcoyne (approved):

Issue #2674, opened 2024-12-05
Request to change one exhibit slug


caaster:
Here is the background re my request. I am building a Looker studio list that leverages GA4 -- to gather high-level stats for each published exhibit. I ran into problems adding some exhibits to the Looker template yesterday, as described below. Would it be possible (see Jacob's comments) to change the slug `blrcc` to `blrcc1`?

Hi Jacob, Something strange has happened with the [Looker template](https://lookerstudio.google.com/u/0/reporting/8b40d476-0e79-4fae-93fa-122b8e9b777b/page/p_ydvl4zabnd/edit). I have successfully added almost all exhibits, except the following 6 that I added are not appearing in the Looker spreadsheet. Can you please reality check me to see if I have done something wrong? I've tried to delete and re-add them to no avail. TIA.
https://exhibits.stanford.edu/blrcc
https://exhibits.stanford.edu/crrw
https://exhibits.stanford.edu/harrison
https://exhibits.stanford.edu/activism
https://exhibits.stanford.edu/publicsyncopations
https://exhibits.stanford.edu/mednarratives

Jacob Hill
  9:25 AM
[...]

PR #2765, opened 2024-12-18
fix focus ring, #2674

### LINKED ISSUE #2764: Browse category and Feature page thumbnail focus ring
There is an issue with the focus ring not appearing correctly on the new Browse category and Feature page thumbnail cards. Compare with the old card focus ring.

Browse category card example: https://exhibits.stanford.edu/martin-wong/browse
Feature page card example: https://exhibits.stanford.edu/virtual-tribunals

https://github.com/user-attachments/assets/21e3eb2a-1c85-4ba9-8b64-2c924ed78e6b

https://github.com/user-attachments/assets/b566efe5-087b-4e65-b542-2e78f05fbe9c

dnoneill:
closes #2764

Screenshot 2024-12-18 at 2 19 06 PM

### DIFF
--- a/app/assets/stylesheets/modules/accessibility.scss
+++ b/app/assets/stylesheets/modules/accessibility.scss
@@ -5,3 +5,9 @@
     outline: $input-focus-border-color auto 5px;
   }
 }
+
+/* Needed for box-shadow to wrap around the entire element */
+/* Component library uses box-shadow to provide a two color focus ring */
+.browse-landing a, .browse-categories a {
+  display: inline-flex;
+}


corylown (commented):
@dnoneill it looks like this problem is getting introduced by the component library -- something in accessibility.css maybe? Focus ring works as expected upstream in spotlight. Is there a fix we can make to the component library?

app/assets/stylesheets/modules/accessibility.scss
```diff
@@ -5,3 +5,8 @@
     outline: $input-focus-border-color auto 5px;
   }
 }
+
+/* Needed for focus ring to wrap around the element */
+.browse-landing a {
```
corylown, line 10: We also need this applied to `.browse-categories a` for widgets that embed browse categories, for example: https://exhibits.stanford.edu/richard-weiland/feature/ric-s-story

corylown (changes requested):
@dnoneill after poking at this a bit I agree this is a tricky problem. I'm willing to merge this as a fix until we can come up with something different. For now, could you add a comment to the stylesheet explaining that the component library's use of box-shadow to add a second color focus ring doesn't work for this case so this workaround is needed?

corylown (approved):

PR #2780, opened 2025-01-06
Test with Ruby 3.4

jcoyne:


jcoyne:
It seems like webmock and http.rb and ruby 3.4 do not work together.  I believe http.rb is pulled in via traject.

PR #2786, opened 2025-01-07
Swap order of contents and sidebar

### LINKED ISSUE #2466: Keyboard focus order for Sidebar and Search results should match visual order
### Current behavior
The programmatic order (DOM, focus order) does not match the visual order.

In DOM-order, the _Limit your search_ filters ("sidebar" section) comes after _Search results_ ("content" section), but appears before the _Search results_ in visual L-R reading order.

See the following WCAG techniques:
- [G59: Placing the interactive elements in an order that follows sequences and relationships within the content](https://www.w3.org/WAI/WCAG21/Techniques/general/G59)
- [C27: Making the DOM order match the visual order](https://www.w3.org/WAI/WCAG21/Techniques/css/C27)

### Steps to reproduce

WCAG Level A: [Understanding Success Criterion 2.4.3: Focus Order](https://www.w3.org/WAI/WCAG21/Understanding/focus-order.html).

Example page: [Parker Library - Search results - List view](https://exhibits.stanford.edu/parker/catalog?f%5Bformat_main_ssim%5D%5B%5D=Archive%2FManuscript&search_field=manuscript_number&sort=title_sort+asc%2C+pub_year_isi+desc&view=list).

Use keyboard Tab/Shift+Tab to navigate the page. Traverse all _Search results_ before keyboard focus lands on the _Limit your search_ filters.

### Possible solution

Switch the DOM position of `<section id="contents">` and `<section id="sidebar">` so that the sidebar filters precede the search results.

**Note**: The same "sidebar" and "content" arrangement is part of Blacklight and appears in other applications, including [Arclight Beta](https://archives.stanford.edu/collections). The same issue is already corrected in Argo: https://github.com/sul-dlss/argo/issues/4161.

taylor-steve:
Closes #2466.

Swaps the order of content and sidebar so that the facets precede content in the tab order.

Note if you test with `annotate_rendered_view_with_filenames` enabled, that counts as content for `content_for?` so there will be an empty sidebar.

### DIFF
--- a/app/views/layouts/spotlight/spotlight.html.erb
+++ b/app/views/layouts/spotlight/spotlight.html.erb
@@ -1,0 +1,20 @@
+<%# Override Spotlight layout for accessibility: https://github.com/sul-dlss/exhibits/issues/2466 %>
+<%# TODO: Spotlight v5 changes this layout %>
+
+<% content_for(:content) do %>
+  <% if content_for? :sidebar %>
+    <aside id="sidebar" class="<%= sidebar_classes %> <%= content_for(:sidebar_position) || 'order-first' %>" aria-label="<%= t('blacklight.search.documents.aria.limit_search') %>">
+      <%= content_for(:sidebar) %>
+    </aside>
+
+    <section id="content" class="<%= main_content_classes %> <%= content_for(:content_position) || 'order-last' %>" aria-label="<%= t('blacklight.search.search_results') %>">
+      <%= yield %>
+    </section>
+  <% else %>
+    <section id="content" class="col-md-12">
+      <%= yield %>
+    </section>
+  <% end %>
+<% end %>
+
+<%= render template: "layouts/spotlight/base" %>


taylor-steve:
@dnoneill That seems likely, but I don't recall. Do you have any more specifics? We should add that context to the ticket. Did we intend to get rid of the Spotlight specific page layout in Spotlight?

app/views/layouts/spotlight/spotlight.html.erb
```diff
@@ -0,0 +1,20 @@
+<%# Override Spotlight layout for accessibility: https://github.com/sul-dlss/exhibits/issues/2466 %>
+<%# TODO: Spotlight v5 changes this layout %>
```
jcoyne, line 2: Can we upgrade to Spotlight 5 first?  This seems to be putting obstacles in our way for future maintenance.

taylor-steve: Most likely? I don't think this is very urgent. Anyway, it's a 4-character change that doesn't impact production usage so not particularly scary as far as obstacles go.

dnoneill:
There isn't a spotlight specific layout? I am confused.

app/views/layouts/spotlight/spotlight.html.erb
jcoyne: It's not so much that it's a 4 character change. It's adding a new file that we need to remove and test at some later date.

taylor-steve: What file would need to be removed? This file would remain.

jcoyne: Doesn't this override what is already provided by spotlight?  If you override stuff from upstream, you can drift from the upstream, making maintenance more complicated.

taylor-steve: I am unsure of what your concern is here. Are you suggesting that we shouldn't override this layout at all to implement the requested change? Or are you taking issue with the timing of the change? I assumed the latter. I'm saying this file works for Spotlight v5 as intended with no intervention. However, there's a QOL helper name change in the Spotlight v5 layout (`content_for?` to `content?`), so I put a reminder in. For a draft PR.

jcoyne: I'm suggesting that we shouldn't do any overrides that aren't strictly necessary.  If we can upgrade versions first and avoid doing this override, then we should do that unless there are compelling reasons not to.

taylor-steve: I see. There is no current structural change in this layout between Spotlight v4 and v5. Exhibits would need this specific override either way. The alternative would be to swap the order in Spotlight.

The order swap decision was contentious, so I didn't know if we were going to impose that for all Spotlight users by default (I think that's what @dnoneill is getting at, that maybe we already decided to do that).

jcoyne: "The order swap decision was contentious"

Can you provide more information about this?  I'm wondering if we can spend more time building consensus in the community rather than diverging without having tried to work this out.

taylor-steve: Did you read the issue and the linked work within?

jcoyne: Yes, and I didn't see anything contentious in there.

jcoyne: It seems like we should fix this in Blacklight(https://github.com/projectblacklight/blacklight/issues/3252)  as it affects all Blacklight projects, not just exhibits.

taylor-steve:
Closing in favor of https://github.com/projectblacklight/spotlight/pull/3405

PR #2787, opened 2025-01-07
convert to jsbundling

jcoyne:


jcoyne:
Currently blocked by https://github.com/sul-dlss/blacklight_heatmaps/pull/82

PR #2809, opened 2025-01-15
have all navigation links appear in mobile menu, move mobile menu to …

### LINKED ISSUE #2772: Header: All menu navigation links should appear in the hamburger dropdown
Currently, only the top-level navigation links appear in the hamburger. The lower-level links appear below the title.
https://exhibits.stanford.edu/martin-wong
Image

We'd like all of the navigation links to appear in the hamburger. As an example, see the [Archives implementation](https://archives.stanford.edu/collections).
Image

This is already implemented in the component library. See example: https://sul-dlss.github.io/component-library/header/.
Image

dnoneill:
…underneath title

closes #2772 
Screenshot 2025-01-15 at 5 41 54 PM
Screenshot 2025-01-15 at 5 41 46 PM

### DIFF
--- a/app/assets/stylesheets/component_library_overrides.css
+++ b/app/assets/stylesheets/component_library_overrides.css
@@ -21,3 +21,8 @@
 .navbar-nav {
   --bs-nav-link-color: black;
 }
+
+/* makes the hr span the length of the screen */
+.navbar-collapse hr {
+  margin-right: -100vw;
+}
--- a/app/components/exhibit_navigation_component.html.erb
+++ b/app/components/exhibit_navigation_component.html.erb
@@ -2,7 +2,7 @@
   <%= link_to(helpers.current_exhibit.title, helpers.spotlight.exhibit_path(helpers.current_exhibit), class: 'navbar-brand') %>
 <% end %>
 
-<ul class="masthead-navigation d-flex flex-row flex-wrap justify-content-end list-unstyled mb-0 align-self-end">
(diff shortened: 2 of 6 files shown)


corylown:
@dnoneill I'm seeing that the utility links are wrapping below the subtitle when a subtitle is present. I think the design wants the utility links to be inline with with subtitle if it's present or the title if the subtitle is not present.
Screenshot 2025-01-16 at 9 16 51 AM
Screenshot 2025-01-16 at 9 16 34 AM

corylown:
@dnoneill no matter what I try the utility links always wrap to the next line. When you're back let's take a look at this together.

app/assets/stylesheets/component_library_overrides.css
```diff
@@ -21,3 +21,7 @@
 .navbar-nav {
   --bs-nav-link-color: black;
 }
+
+.dropdown-menu li {
+  display: block!important;
```
jcoyne, line 26: Would you mind leaving a comment here about why we need the `!important` directive? also, please add a space in between.

jcoyne (changes requested):

app/assets/stylesheets/component_library_overrides.css
dnoneill: This has been hanging out so long Nick has fixed it in the component library, so I removed the code. It won't be an issue once a new release is cut and for now I think it's fine if the dropdown appears horizontally, especially since I think we will get a new release before this goes into production.

PR #2814, opened 2025-01-22
Disable rspec monkeypatching

jcoyne:


corylown (commented):
There are still a few left to convert: PurlService, PurlModsService, and PurlCollectionMemberDruids

PR #2821, opened 2025-01-24
Update dependencies; update spotlight to 4.7

corylown:


package.json
```diff
@@ -14,7 +14,7 @@
     "postcss": "^8.4.49",
     "postcss-cli": "^11.0.0",
     "sass": "^1.82.0",
-    "spotlight-frontend": "^4.5.0"
+    "spotlight-frontend": "^4.7.0"
```
jcoyne, line 17: Shouldn't blacklight-frontend also be updated?

jcoyne (changes requested):

package.json
corylown: Yes, thank you. Done.

jcoyne (approved):

PR #2831, opened 2025-01-29
Extract a component for embedded mirador

jcoyne:
This helps move code out of erb templates.


dnoneill (changes requested):
Screenshot 2025-01-29 at 5 09 20 PM

app/components/custom_viewer_component.html.erb
```diff
@@ -1,5 +1,5 @@
 <% if document.external_iiif? %>
-  <%= render partial: "embedded_mirador3", locals: {document: document} %>
+  <%= render EmbeddedMirador.new(document:) %>
```
dnoneill, line 2:
```suggestion
  <%= render EmbeddedMirador.new(document:) %>
  <%= render EmbeddedMiradorComponent.new(document:) %>
```

jcoyne:
@dnoneill thank you for catching that. I added a test for `CustomViewerComonent`, because it seems like that path was totally without tests.

dnoneill (approved):

PR #2834, opened 2025-01-30
Increase subheading size

jcoyne:
And add other bs classes to subheading
See https://github.com/projectblacklight/spotlight/pull/3374


taylor-steve:
This seems to break https://github.com/sul-dlss/exhibits/pull/2767 for me. I can poke at this if you don't have time.

PR #2835, opened 2025-01-30
Increase subheading size

taylor-steve:
@jcoyne found and mentions where this changed in #2834 but given https://github.com/sul-dlss/exhibits/pull/2767 we should omit the display classes.


jcoyne:
Is this a duplicate of #2834 ?

jcoyne (approved):

PR #2838, opened 2025-01-31
Seperate controller

dnoneill:
@corylown ```copy_blacklight_config_from(CatalogController)``` gave that same header error. Went with the configurable option.


app/models/viewer.rb
```diff
@@ -11,7 +11,7 @@ class Viewer < ApplicationRecord
             allow_blank: true
 
   def default_viewer_path
-    'oembed_default'
+    'catalog/oembed_default'
```
dnoneill, line 14: Pretty sure this doesn't break anything. I tested to make sure but I could have missed somewhere.

PR #2845, opened 2025-02-06
Add image selection for item embed viewers

### LINKED ISSUE #2604: Enable image area selection for Item Embed - Mirador

taylor-steve:
Closes #2604.

Continuation & partial refactor of the work done in https://github.com/sul-dlss/exhibits/pull/2752.
Still needs https://github.com/sul-dlss/sul-embed/pull/2279 to work.

This can be run locally. For example if sul-embed is running at localhost:3001:
```rb
# config/initializers/oembed.rb
purl_provider = OEmbed::Provider.new('http://localhost:3001/embed.json?&hide_title=true')
```

https://github.com/projectblacklight/spotlight/issues/3431 causes a problem when there are multiple copies of the same item in a single embed block. In that case, the panels have duplicate IDs and the first panel will always get the image selection update regardless of which panel the area selection was initiated from. Will need to fix that first.

Changes from https://github.com/sul-dlss/exhibits/pull/2752:
* Isolates widget modification & viewer state sync to their own classes
* Removes the mutation observer in favor of specific events
* Reduces the scope of the SirTrevor overrides
* Removes the wildcard origin in postMessage
* Syncs the state between the existing Spotlight "multi-image selector" page selection and the viewer


app/components/custom_viewer_component.rb
```diff
@@ -5,12 +5,19 @@
...
     super
 
     @document = document
     @presenter = presenter
     @view_config = view_config
-    @block_context = block_context
+    @block_context = correct_block(block_context)
+  end
+
+  def correct_block(block_context)
```
corylown, line 19: The name `correct_block` seems vague. It seems to be selecting/creating a block based on the document id, so maybe `block_for_document` or something?

corylown (approved):

PR #2917, opened 2025-05-29
Add job to import alt text spreadsheet

taylor-steve:
This adds a job to import the AI Alt Text Experiment spreadsheets.

### Why are we doing this?
In summary I was told:
* We need a way to import these sheets
* A developer will be doing the imports, not the end-users
* We might do these imports periodically for a while, but if AI alt text proves useful we would be moving away from spreadsheets to some more streamlined (in-app?) process.

Given that, and that I don't know of any other good ways to inject Sir Trevor data, I felt a self-contained job would be a fine method.

### What does this actually do?
In short it:
* tries to pluck an ID from the image URL from the sheet (e.g., druid for solr docs, relative URL for uploaded images)
* tries to find the exhibit and the page
* looks at every item in each Sir Trevor block on the page and matches the image ID
* If there is a singular match, it updates the alt text for that particular image based on the spreadsheet logic

### Notes
* This doesn't cover all block cases. For example, Spotlight/Exhibits *can* use the solr document blocks with uploaded files but none of the current spreadsheets are targeting this case. Nothing bad would happen if somebody ran one later, it would fail to match.
* Manipulating the Sir Trevor data is not fun.

### DIFF
--- a/app/jobs/import_page_alt_text_job.rb
+++ b/app/jobs/import_page_alt_text_job.rb
@@ -1,0 +1,220 @@
+# frozen_string_literal: true
+
+# A job to import alt text for home/about/feature pages from the AI alt text experiments.
+class ImportPageAltTextJob < ApplicationJob
+  def perform(csv_path:, dry_run: false)
+    update = !dry_run
+
+    CSV.foreach(csv_path, headers: true) do |csv_row|
+      row = AltTextSpreadsheetRow.new(csv_row)
+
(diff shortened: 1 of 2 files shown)


jcoyne:
Do we really want to merge this? It feels like a one time thing.

taylor-steve:
Yeah I think it's fine not to merge. I have no idea what is happening next.

PR #2929, opened 2025-06-13
Rm collection indirection

corylown:
We were creating an object we don't need.


app/services/purl_collections.rb
```diff
@@ -0,0 +1,45 @@
...
+  # @return [Array<Purl>] array of Purl objects for the collections this Purl belongs to
+  def collections
+    @collections ||= collection_druids.map { |druid| Purl.new(druid) }
+  end
+
+  private
+
+  def collection_druids
+    predicate_nodes.reject { |node| node.value.empty? }.map do |node|
+      node.value.split('druid:').last
```
thatbudakguy, line 29: I have been doing `delete_prefix('druid:')` for this bc it feels slightly more idiomatic, fwiw

thatbudakguy (approved):

PR #2946, opened 2025-07-02
Use cocina for thumbnails metadata

corylown:
Note this is a pull request on the `cocina` feature branch NOT `main`


thatbudakguy (approved):
I think some of this stuff would make sense to migrate to `cocina_display` eventually, especially the `virtual_object?` helper.

PR #2948, opened 2025-07-10
Folio cocina indexer

corylown:
Note: this is a PR on the cocina feature branch not main.


app/models/purl.rb
```diff
@@ -27,19 +25,14 @@ def public_xml
 
   # @return [Hash] the public cocina hash for this Purl object
   def public_cocina
-    @public_cocina ||= JSON.parse(purl_cocina_service.response_body.presence || '{}')
+    @public_cocina ||= JSON.parse(cocina_service_response_body)
```
thatbudakguy, line 28: I think if you wanted to avoid re-parsing the JSON, you could have this be `cocina_record.cocina_doc`, which is just the underlying hash parsed from the JSON. In that case you might not even need to memoize, since the call to `cocina_record` is itself memoized.

app/services/cocina_physical_location.rb
```diff
@@ -0,0 +1,67 @@
+# frozen_string_literal: true
+
+# Class to handle Cocina physical location information
+class CocinaPhysicalLocation
```
thatbudakguy, line 4: 👍🏻 like this choice to extract this as a class to manage the various accessors. Also glad to see `#path` getting some use here.

app/models/purl.rb
```diff
@@ -124,6 +118,10 @@ def purl_virtual_object
     @purl_virtual_object ||= PurlVirtualObject.new(public_cocina:)
   end
 
+  def cocina_service_response_body
+    @cocina_service_response_body ||= purl_cocina_service.response_body.presence || '{}'
+  end
+
   # Normalize the role text to use consistent capitalization and remove trailing punctuation.
   def format_role(role_element)
     role_element.text.strip.capitalize.sub(/[.,:;]+$/, '').tr('|', '')
```
thatbudakguy, line 127: Is this worth putting into the gem? Presumably we don't want this weird punctuation in any roles, although I didn't see data like this when I was working on it

lib/traject/macros/dor.rb
```diff
@@ -12,13 +12,55 @@ def stanford_mods(method, *args, **kwargs)
...
+      # @param path [String] the JSON path expression to evaluate
+      # @example
+      #   cocina_display_path('$.description.note[?match(@.type, "table of contents")].value')
+      def cocina_display_path(path)
+        lambda do |resource, accumulator, _context|
+          Array(resource.cocina_record.path(path)).each do |v|
+            accumulator << v if v.present?
+          end
+        end
+      end
```
thatbudakguy, line 38: These are pretty nice; I can see putting them into a traject-specific module in the Gem. When I get around to doing Earthworks/Searchworks indexing maybe it'll be time to do that

app/models/purl.rb
thatbudakguy: Oh, gotcha. Makes sense

PR #2980, opened 2025-09-25
Remove fallback to mods endpoint

jcoyne:
There is no sense in falling back, because purl gets the mods from the public xml.


app/services/purl_mods_service.rb
```diff
@@ -34,19 +29,6 @@ def mods_xpath
...
   end
 
-  def mods_response_body
-    PurlService.new(druid, format: :mods).response_body
-  end
-
-  def send_fallback_notification!
-    return unless defined?(Honeybadger)
-
-    Honeybadger.notify(
```
jcoyne, line 44: We never see this notification anymore

PR #2983, opened 2025-10-02
Check response status when indexing

jcoyne:


dnoneill:
I think this can go in but Chris did find the problem file

PR #2989, opened 2025-10-24
remove default turbo.drive enabling

### LINKED ISSUE #2988: Error handling for duplicate slugs
If the user tries to create an exhibit with an already existing slug there is no information and the console just spits out the below. It is unclear to both user and developer what is going on. We should have an alert that tells the user that the slug already exists. ~~I would check if this is an issue in spotlight first and fix it there if it is.~~ Not an issue in spotlight

Image

https://stanfordlib.slack.com/archives/C0BTKKS8P/p1761243089151609

dnoneill:
closes #2988 

I really don't understand what is going on. This fixes the issue but I don't really understand Turbo drive and could use pairing on this.

### DIFF
--- a/app/javascript/application.js
+++ b/app/javascript/application.js
@@ -1,5 +1,6 @@
 // Entry point for the build script in your package.json
 import "@hotwired/turbo-rails"
+Turbo.session.drive = false
 import "./controllers"
 
 // Required by Blacklight


jcoyne:
The problem is that the rails endpoint to which you are submitting the form to is not doing a redirect as required by Turbo.  I don't think turning Turbo off globally is the best solution.

PR #3023, opened 2025-12-11
Update Spotlight to 5.2.x

taylor-steve:
Need to unravel some dependency resolution issues before we can get to Rails 8.1


Gemfile
```diff
@@ -85,7 +85,7 @@ gem 'bootstrap_form', '~> 5.4'
 gem 'blacklight', '~> 8.0'
 gem 'blacklight-gallery', '~> 4.8'
 gem 'blacklight_heatmaps', '~> 2.0'
-gem 'blacklight-spotlight', '>= 5.0.0.pre.alpha1'
+gem 'blacklight-spotlight', '~> 5.2.0'
```
corylown: Should we change this to `~> 5.2` so we can get minor updates automatically?
```suggestion
gem 'blacklight-spotlight', '~> 5.2'
```

PR #3038, opened 2025-12-19
Permit SiteImprove IPs

taylor-steve:
https://stanfordlib.slack.com/archives/C020B6YQSHE/p1765818055889189


corylown (commented):
@taylor-steve I don't think we need this since we're already allowing the Siteimprove user agent string through: https://github.com/sul-dlss/exhibits/blob/2765e57b9c404d082f0cbdd4d2cb9338e3b4de6b/config/initializers/bot_challenge_page.rb#L47

PR #3048, opened 2026-01-10
Increase Capybara's default_max_wait_time in CI

taylor-steve:


jcoyne:
I still saw the error in dependenecy updates today.

PR #3131, opened 2026-06-01
Update spotlight and devise

jcoyne:


corylown:
These were updated by https://github.com/sul-dlss/exhibits/pull/3132

PR #3156, opened 2026-08-07
update component library

dnoneill:


corylown (changes requested):
The search bar isn't quite right on both the landing page and exhibit pages.

Landing page on main:
Image

Landing page on component-library branch:
Image

(The search bar is using up more horizontal space there's too much space taken up by the search dropdown on the left.)

Exhibit page on main:
Image

Exhibit page on component-library branch:
Image

(The search field labels are getting cut off because there's not enough space)

